PR #17, opened 2015-03-21
Add xsv, byte swapping, cbor, suffix arrays, regexes and elastic tabstops

BurntSushi:
I also added a new category "text processing" with three entries because
they didn't really seem to fit any where else.


kud1ing:
Awesome, thanks.

PR #27, opened 2015-05-07
Add rust-compress, blas, lapack.

cmr:


kud1ing:
Thanks, i will remove compress though, because of "This is not a production-quality library, it is a proof of concept."

PR #40, opened 2015-05-15
Alternative MessagePack serialization library.

3Hren:
This library supports both low-level and high-level interface allowing to build either performance-critical or convenient applications. It also compiles with the latest Rust channel and will continue to do so.


kud1ing:
Thanks

PR #55, opened 2015-07-05
list serde-rs, a framework to generically serialize Rust data structures

fkr:
Hi,

serde-rs is pretty popular framework for serialization of data structures.
Am not sure wether I've added it to the right category - but this seems to be the best fit.

regards,

felix


kud1ing:
I am not sure either. Could also fit "encoding". Merged it as is for the time being.
Thanks

PR #58, opened 2015-07-20
Add rust-asn1

frewsxcv:


kud1ing:
Thanks

kud1ing:
Hm, is this really cryptography specific or would it fit "Encoding" ?

frewsxcv:
Ah yes, you're right

kud1ing:
Ok, I've moved it and added the Travis badge

PR #61, opened 2015-08-01
Elasticsearch is still called Elasticsearch

skade:
The database is Elasticsearch, only the company changed name to Elastic.


kud1ing:
Makes sense, thanks

PR #67, opened 2015-08-14
Added RustCamp 2015 talks to resources

0xCMP:
I thought of making a videos or conference section, but I think once there are more conferences or known video series then a section for those will make sense.


kud1ing:
Thanks

PR #72, opened 2015-08-28
add mrhooray/torch under profiling

mrhooray:


kud1ing:
This is awesome, thanks.

PR #83, opened 2015-10-18
Add awesome list badge

umcsdon:
https://github.com/sindresorhus/awesome/issues/253


kud1ing:
Thanks, but at this point of time i won't merge something like this.
See also https://github.com/kud1ing/awesome-rust/pull/64

PR #92, opened 2015-11-06
Added arrayfire-rust, inquerest and queryst

ivanceras:
arrayfire-rust - rust binding for arrayfire which uses opencl underneath the hood
inquerest and queryst - url param parsers


kud1ing:
Thanks

PR #96, opened 2015-11-19
Add my `ngrams` crate to "Text Processing"

pwoolcoc:
I put this under Text Processing because, even though the library can be used with arbitrary iterators, n-grams are most often used with sequences of text


kud1ing:
Thanks

PR #98, opened 2015-11-21
add Authentication category and rust-jwt

pyros2097:


kud1ing:
I am wondering, whether that should be a section in "Web programming". What do you think?

pyros2097:
Yeah I guess it could be. I was just following the every other awesome list like ruby, python, go all of them have it under Authentication and OAuth. Its your call.

kud1ing:
Merged it as is. Will think about it. Thanks.

PR #99, opened 2015-11-29
add resource embedding section and rust-embed

pyros2097:


README.md
```diff
@@ -402,6 +403,8 @@ See also [http://areweideyet.com/](http://areweideyet.com/) and [Rust and IDEs](
 * Windows
   * [retep998/winapi-rs](https://github.com/retep998/winapi-rs) — Windows API bindings [<img src="https://travis-ci.org/retep998/winapi-rs.svg?branch=master">](https://travis-ci.org/retep998/winapi-rs)
 
+### Resource Embedding
+* [pyros2097/rust-embed](https://github.com/pyros2097/rust-embed]) — Generates rust code to embed resource files into your rust executable
```
CryZe: There's a trailing ] in your URL

CryZe:
Also what's the purpose of this when there's [!include_bytes](http://doc.rust-lang.org/std/macro.include_bytes!.html) and [!include_str](http://doc.rust-lang.org/std/macro.include_str!.html)?

pyros2097:
Oops typo. Well I was learning rust and decided to do this as a learning practice. Man! I thought there should be a possible way to include files using macro. But this maybe has better use case as it can dynamically also load files(todo) for debugging purposes maybe and is useful for serving files through http.

suhr:
Actually, it would be nice to have a wrapper around `include_bytes!` and `include_str!`, so you could do something like `asset!("data/foo.bin")` or `asset_str!("styles/bar.css")`.
This should be implemented as a compiler plugin though.

pyros2097:
@suhr Thanks I'll try to figure that out --> using compiler plugins that is

pyros2097:
Done! Anyone want to look at to.
https://github.com/pyros2097/rust-embed
I've implemented it as a compiler plugin/macro
```rust
let asset = embed!("examples/public".to_owned());
asset("/index.html".to_owned()).unwrap();
```

PR #125, opened 2016-02-07
Added Tokei

Aaronepower:
[Tokei](https://github.com/Aaronepower/tokei), is a program some people might want, maybe?


kud1ing:
Who could live without it. Thanks.

PR #127, opened 2016-02-12
Updated link to the list of awesome awesomeness to point to the most …

bigtunacan:
…currently maintained list as the one listed here is out of date and has not been maintained in a long time.


kud1ing:
Thank you. I've decided to not link to any particular meta-awesome document at all. The concept is probably understood by now.

PR #129, opened 2016-02-24
Remove `cargo-fmt`

pwoolcoc:
The `fmt` subcommand was moved to the `rustfmt` project, and the `pwoolcoc/cargo-fmt` project is no longer active.


kud1ing:
Thank you

PR #146, opened 2016-04-14
Add systemd-manager

mmstick:
A while back I wrote an application for managing systemd services on Linux systems with Rust and GTK-rs for the GUI front-end.


kud1ing:
Thank you

PR #149, opened 2016-04-20
Add an .md file that attach stars count and last updated date

chaconnewu:
Added a separate file that is a complete clone of the original README.md file, attaching the number of stars as well as last pushed date for each GitHub repositories included in the curated list. The date was collected through GitHub API (just now). And I can update the file regularly or as needed.

Let me know if you find it interesting. Thanks!

View the fille directly at:
https://github.com/chaconnewu/dashboard_for_awesomes/blob/master/awesomes/awesome_rust_with_repo_info.md

### DIFF
--- a/awesome_rust_with_repo_info.md
+++ b/awesome_rust_with_repo_info.md
@@ -1,0 +1,582 @@
+Awesome Rust
+============
+
+A curated list of Rust code and resources, inspired by other awesome lists. The goal is to have only projects that are mostly stable and useful to users.
+
+If you want to contribute, please read [this](CONTRIBUTING.md) .
+
+-   [Awesome Rust](#awesome-rust)
+    -   [Applications written in Rust](#applications-written-in-rust)
+        -   [Games](#games)
+        -   [Operating systems](#operating-systems)
+        -   [System tools](#system-tools)
+    -   [Development Tools](#development-tools)
+        -   [Debugging](#debugging)
+        -   [Embedded](#embedded)
+        -   [FFI](#ffi)
+        -   [IDEs](#ides)
+        -   [Profiling](#profiling)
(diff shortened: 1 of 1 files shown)


kud1ing:
Nice scripting there. :) An overview of such stats is certainly interesting.

OTOH i find it's in the nature of a project like this that its content tends to overwhelm. We have many entries. I'd like to keep each entry as minimalistic as possible because i fear adding to each of them increases the debt of overwhelm.

Thank you, and keep up the good work.

chaconnewu:
Hi @kud1ing, thanks a lot for the feedback. I want to point out that this is a separate file, which will not overwrite your existing one. So users will have a minimalistic experience by default. If they want more info, they can browse the one I generated. What do you think?

kud1ing:
Yes, i understand.

In the beginning i'd generated the markdown file from a Python script. That made sure that format was correct everywhere and that you could have "duplicate" entries in different places without actually duplicating the information. It turned out that people assumed markdown files to be a source and not a product. So this caused some problems.

I feel that the informational scope of this project should be kept small. But it think there is a lot of room for interesting projects like yours for mixing in additional (live) data.

chaconnewu:
Thanks a lot for your response @kud1ing! Sorry that I have many questions :). But I'm a HCI researcher, trying to understand and improve curation practice (my profile: https://scholar.google.com/citations?user=1TNPqLMAAAAJ&hl=en). So I'm very interested in your experience & perception of the curation practice as well. 

I'm wondering what did you mean by "people assumed markdown files to be a source and not a product"? And what problems it caused?

You are definitely free to answer me or not.

kud1ing:
What i meant is, that people had send pull requests where the generated markdown file was modified, instead of the script.

PR #152, opened 2016-05-11
Add tv-renamer

mmstick:
A new application that I've written which features an optional GTK3 user interface and TVDB support for adding titles to episodes.


kud1ing:
Thank you

PR #158, opened 2016-05-22
Add github/contain-rs collection

pravic:
Rust's std::collections extensions and new useful libraries like trie, lists, lru-cache and so on.


kud1ing:
Thanks

PR #160, opened 2016-05-22
Add unicorn-rs: Rust bindings for the unicorn CPU emulator

pravic:
I am not sure which section more suitable — FFI (because it is bindings) or Virtualization (because its a CPU emulator bindings).


kud1ing:
Yes, it's not always easy. Thank you.

PR #167, opened 2016-06-16
Add 'Awesome Rust @LibHunt' to the 'Resources' section

StanBright:
This is the Awesome Rust list with ordering by popularity, dev activity and additional metadata. Moreover, the site offers comparing different libs from the same category.

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,6 +655,7 @@ See also [Rust web framework comparison](https://github.com/flosse/rust-web-fram
 * [Rust Guidelines](http://aturon.github.io/)
 * [rust-learning](https://github.com/ctjhoa/rust-learning) — a collection of useful resources to learn Rust
 * [Rustlings](https://github.com/carols10cents/rustlings) — small exercises to get you used to reading and writing Rust code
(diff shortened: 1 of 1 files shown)


kud1ing:
Thank you.

As i understand it, the community curates that data and you keep both the data and the advertisment money. Is that right?

In this case i have to decline. Feel free to clarify and reopen.

StanBright:
@kud1ing as the owner of the repo, you have all the rights to do whatever you want. I'm not sure what I have to clarify...

"advertisement money"... Last month, I got "$55.09". I think, that's not too much for the time I've invested... even if it was more, I don't think that would be bad. Especially, if that's because you've built something useful.

Away from that, rust.libhunt.com offers added value to the list at GitHub. The option to view projects' popularity and dev activity is really useful when comparing libs. As a dev myself, I find it helpful, and I believe others will do, too. 

Also, I'm not sure you've noticed that the list here is mentioned and linked as "the official list" (at the contribute page).

Anyway, thumbs up for maintaining "Awesome Rust" here! I understand that you are trying to keep it clean and useful.

PR #170, opened 2016-06-29
Added cargo-modules

regexident:


kud1ing:
Thanks. Except for the second difference where 2 crate links are removed.
Was that intentional?

regexident:
> 2 crate links are removed in the second difference. Was that intentional?

Well yeah. Kinda. Given that none of the other cargo plugins had those second links I though those probably were redundant remnants of previous edit, or something, not noticing that the links differed.
Anyway, I fixed it and added the missing crate links for all other cargo plugins.

kud1ing:
Thank you

regexident:
Also, ei guude wie! ✋  😃  Have we met yet?
Either way you should really show up at [Rust Rhein-Main](http://meetup.com/Rust-Rhein-Main) some time! 👍

kud1ing:
:)
Yes, i've considered the gathering in Textor, but couldn't make it.
Next time.

PR #177, opened 2016-07-26
Add quantum project

beneills:


kud1ing:
Hi, thank you.

I know the hierarchical grouping is not perfect.
I am on phone right now so my impression might not be accurate. Would you think that your link would also fit under "Applications written in Rust > Emulators"?

beneills:
Hi,

I'd say it is an emulator, although it's certainly a library crate and not
an end user application.

I'm happy to put it anywhere!

Ben

kud1ing:
Merged as is. Thank you.

PR #193, opened 2016-08-25
Add andschwa/rust-genetic-algorithm

andschwa:
I wrote this Rust application in May 2015 to learn Rust and practice my GA skills. It could easily be rewritten as a library (the GA part is already), and I believe would be useful as example code. While developing, I worked closely with the Rust IRC community to produce correct, idiomatic code.


kud1ing:
Thank you

Issue #197, opened 2016-09-08
Incorrect game engine entry


ebkalderon:
As of early 2016, Amethyst is no longer a lone-developer project. :smile: As such, the new GitHub repository is located at http://github.com/amethyst/amethyst, the website is https://www.amethyst.rs/, and the correct Travis CI shield is at https://travis-ci.org/amethyst/amethyst.svg?branch=master.

Also, in regards to `rust.libhunt.com`, which I assume this repository is affiliated with, I don't think there should be a `Game Development/Amethyst` sub-category. The only entries in there are `amethyst` and an unrelated crate called `challonge-rs`.

I'm willing to make a PR to fix these issues. That said, this is an excellent list! Keep up the great work.

kud1ing:
Thank you for the kind words and taking the time to improve this list. But first and foremost thanks for writing interesting Rust code. Without it this list would be empty. :)

kud1ing:
This list is not affiliated with rust.libhunt.com. They are importing the data from this list, but afaik keep the data that is submitted there for themselves.

PR #198, opened 2016-09-08
Correct game development section

ebkalderon:
This pull request should fix issue #197. Please let me know if there is anything else in this section you would like fixed.


kud1ing:
Thank you

PR #200, opened 2016-09-12
Adds deck link for Learning systems programming with Rust.

jxson:


kud1ing:
Thank you. TIL what a deck is.

PR #211, opened 2016-10-16
Remove tangle

tmccombs:
Because it is deprecated in favor of futures-rs


kud1ing:
Good catch, thank you

PR #231, opened 2016-12-10
Adding rust vim plugins to IDE section

salpalvv:
These are the rust-related plugins I am currently using for development.


kud1ing:
Thank you

PR #238, opened 2017-01-01
Add vscode-rust fork of RustyCode

Timmmm:
It has a 35 commits after RustyCode so I think it is fair to include it.


kud1ing:
Thank you

PR #241, opened 2017-01-08
Added Way Cooler to README

Timidger:
Wasn't sure where to put it, since it doesn't really fit in under the categories. Made a new category "Window Managers" since I could foresee others going here as well.

Please tell me if there is a better place to put it.


kud1ing:
That's a good question. I merge as it is for the time being.
Thank you.

PR #242, opened 2017-01-12
Add Hound

ruuda:
Hound is a WAV encoding and decoding library. It is actively maintained and it has been stable for over a year.


kud1ing:
Thank you

PR #247, opened 2017-01-22
Add Snatch

k0pernicus:
Add as new entry _Snatch_, a simple and fast download accelerator


kud1ing:
Hi, thank you.
Snatch is already there, under "system tools".

PR #248, opened 2017-01-23
Add stdx

brson:
I think stdx is its present form is an excellent learning tool, and this is the best place I see to link it right now.


kud1ing:
Thank you.

PR #250, opened 2017-01-30
Add config-rs

mehcode:
Config is a layered configuration system for Rust applications (with strong support for 12-factor applications).

https://github.com/mehcode/config-rs


kud1ing:
Thank you.

PR #253, opened 2017-02-09
add fs_extra

webdesus:
**fs_extra** - expanding opportunities standard library std::fs and std::io.

**Key features:**
- Copy files with recept information about process.
- Copy directory recursively and recept information about process.
- Copy list directoies and files recursively and recept information about process.
- Move files with recept information about process.
- Move directories recursively and recept information about process.
- Move list directoies and files recursively and recept information about process.
- One method for create and write String content in file.
- One method for open and read String content from file.
- Get size folder


kud1ing:
Thank you.

PR #265, opened 2017-03-01
Added new section in Emulators

ioncodes:
Added Emulator development section for ROM hacking tools, etc and added one for the SNES.


kud1ing:
Thank you.

PR #269, opened 2017-03-17
Update Rust extensions for VSCode.

vadimcn:
Add a note that RustyCode in unmaintained.
Change links to point to VSCode marketplace.


kud1ing:
Thank you.

PR #270, opened 2017-03-18
Updated zip library from slackito/zip to mvdnes/zip-rs

crazymerlyn:
slackito/zip seems to be abandoned (no update in 2 years).
mvdnes/zip-rs is the most used zip crate on crates.io.


kud1ing:
Thank you.

PR #271, opened 2017-03-20
Added arewegameyet.com to GameDev in README

adiba:
The important part is actually the well structured compilation of crates: http://arewegameyet.com/#eco


kud1ing:
Thank you.

PR #276, opened 2017-04-11
added scripting section

fneddy:
added dyon link
added gluon link
added ketos link


kud1ing:
Thank you, that's agood idea.

PR #279, opened 2017-04-16
Add mpw-rs to Cryptography

lispyclouds:
Not sure if this is the right place as there are no Password Managers section.


kud1ing:
Thank you. Yes, i am not sure myself. Let's have it here for the time being.

lispyclouds:
Thanks!

PR #291, opened 2017-05-05
Add whatlang-rs entry

greyblake:
Add entry for [whatlang-rs](https://github.com/greyblake/whatlang-rs)  - natural language detection library


greyblake:
Merci!

PR #298, opened 2017-05-29
Add cargo-find

Ralvke:


Manishearth:
Note: cargo search already does the basics here

PR #304, opened 2017-06-08
Add fd, an alternative to find

crazymerlyn:


luciusmagn:
Nice addition, thanks

PR #305, opened 2017-06-09
Add Zopfli to Compression

mmstick:
Surprised that this wasn't already on here. I've been using this for native Gzip compression since it released last year.


booyaa:
Thanks!

PR #310, opened 2017-06-20
Improve CI

ReadmeCritic:
Looks like crates.io is misreported as an error, white listing it


ReadmeCritic:
There are some ssl issues with ethcore and arrayfire, would you like me to white list those as well?

kud1ing:
Thank you.

PR #312, opened 2017-06-26
Add cbindgen

jrmuizel:
cbindgen is an alternative to rusty-cheddar that doesn't use syntex and is used for generating c bindings for WebRender as used by Gecko


kud1ing:
The URL has a superfluous "r": https://github.com/rlhunt/cbindgen/r

kud1ing:
Added in https://github.com/rust-unofficial/awesome-rust/commit/2d973ca1889f74b101e4f05eca0b432cb84ce4b3

PR #313, opened 2017-06-27
Adding quick-skeleton

arthrp:
Haven't found any criteria to meet to be included in the list, so will try with my small project. I was searching for a scaffolding tool, but found only few ones, which were not so easy to use (in my opinion). So I decided to make my own. It's roughly the same as Yeoman, but simpler. It doesn't have any generators, instead it uses templates (Handlebars) that are filled with user input.


kud1ing:
Thank you. Would you mind moving it above "yaa110/rubigo"?

kud1ing:
I reorder it myself.

arthrp:
Sorry, was away for a few days. Thank you for reordering!

kud1ing:
NP. Thanks for the addition.

PR #315, opened 2017-07-06
Add Mockito to the Testing group

lipanski:
Hello

Mockito has been stable for a while now and I've added some nice features along the past year.
Would be great to see it on awesome-rust :)


kud1ing:
Thank you.

PR #317, opened 2017-07-10
Add rust-djangohashers and sort Cryptography

Racum:
Just adding my little password library, it is very well documented and tested; I think it worth a look. Also, I sorted the Cryptography section.


kud1ing:
Thank you.

PR #319, opened 2017-07-22
Add star count emojis.

d33tah:
This is probably my silliest Github pull request, but I just couldn't help doing it. This PR adds icons that show how many Github stars a project has. The code that added those stars is here:

https://github.com/d33tah/awesome-github-colorer/blob/master/main.py


kud1ing:
Thank you. This looks interesting, but also a bit distracting.
I am glad this can be reapplied using this script to a local copy at any time.

PR #320, opened 2017-08-06
add cargo-update

pahhnv:
- add cargo-update to build system list (subcommand for checking and applying updates to installed executables)
- move cargo-benchcmp to the top of the list (see CONTRIBUTING.md "please pay attention to the alphabetical ordering")
- add build passing/failing badges for cargo-{benchcmp, do, release, script}

Note: cargo-update has been working really good for me on linux, haven't had any problems so far and it definitely has been convenient to use! So I would argue it's both "stable" and "useful" and therefore worthy to be on this list :) At this moment there's appr. 11.5k downloads of cargo-update on crates.io so I think it can even be considered to be "useful to most users", seeing that other cargo subcommands featured here have fewer downloads. This should cover CONTRIBUTING.md.


kud1ing:
Thanks.

PR #321, opened 2017-08-10
Add sharazam/printpdf as PDF library

sharazam:
My PDF library has currently more features than kaj/rust-pdf, which is currently the only listed library. So I thought it would be fair to include it on this list.


kud1ing:
Of course, thank you.

PR #322, opened 2017-08-13
Added Gotham

bradleybeddoes:
We have recently launched [Gotham](https://users.rust-lang.org/t/announcing-gotham/12307) and I wanted to contribute it to the web frameworks list if you'd be so kind as to accept this PR.


kud1ing:
Thank you.

PR #327, opened 2017-08-17
Add Image Processing CLI

spejss:
Hi, I added a link to the image processing CLI that I wrote, which mainly is a wrapper around the Piston image package, but also has implementation for generating histograms.
Since I'm using it at work, maybe it can be useful for others.


kud1ing:
Thank you.

PR #328, opened 2017-08-21
add Module Linker.

fiatjaf:
A Chrome and Firefox extension for browsing source code better. See the GIF below:

http://fiatjaf.alhur.es/module-linker/#/rust


kud1ing:
Thank you.

PR #345, opened 2017-09-26
Add lotabout/skim to the list

lotabout:
Just want to add my crate [lotabout/skim](https://github.com/lotabout/skim) into the list.


kud1ing:
Thank you.

PR #346, opened 2017-10-03
Update README.md

amar-laksh:
ADDED: UTILITIES: workstation - commandline tool to manage your workstations with opencv!


kud1ing:
It's a bit niche, but i've added it anyway.
Thank you.

PR #351, opened 2017-10-05
Try fixing all issues

luciusmagn:
- Fixes a mistake in skim's Travis tag, where the picture was being taken from just `travis-ci.org` while the link lead to `api.travis-ci.org` causing double 302 response.
- Changed the website of thrussh to the github repository, as the original website doesn't exist anymore while the repository still does
- Whitelisted https://sciter.com/ , they are doing something weird with their website, but it still works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: rust
 before_script: gem install awesome_bot
 # `crates.io` needs to be whitelisted, because of https://github.com/rust-lang/crates.io/issues/788
-script: awesome_bot -f README.md --allow-dupe --allow-ssl --white-list crates.io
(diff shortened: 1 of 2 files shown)


dywedir:
(thrussh original repository is https://nest.pijul.com/pijul_org/thrussh)

luciusmagn:
Ah, good to know, although that website times out for me as well. Seems like https://github.com/pijul-scm/thrussh would be the best solution for now

luciusmagn:
@kud1ing are these changes okay? I'd rather not merge something that goes against

kud1ing:
Looks good, thank you a lot.

PR #356, opened 2017-10-12
Add rust-musl-cross to Deployment section

messense:


luciusmagn:
Nice addition, thanks

PR #357, opened 2017-10-12
Add Are we web yet? to Web programming section

tatsuya6502:
- [Are we _web_ yet? -- http://www.arewewebyet.org](http://www.arewewebyet.org)


luciusmagn:
Ah, good idea to add it, too, thanks

PR #358, opened 2017-10-14
Replace snatch with zou

equal-l2:
[snatch](https://github.com/derniercri/snatch) is no longer maintained.
[zou](https://github.com/k0pernicus/zou) is an active fork of snatch.


luciusmagn:
Ah, I see that in this case zou is more or less an official successor with same people involved, so no problem merging. Thanks for updating

PR #361, opened 2017-10-29
 add talks on Rust in Decks & Presentations

rstoke:


luciusmagn:
Nice addition, thanks!

PR #367, opened 2017-10-31
Update vulkano with the new repository.

OakenBow:
The old link leads to a repo that only gives a new link. I also updated the travis-ci to point to the correct project.


luciusmagn:
Thank you

PR #368, opened 2017-11-01
Replace old Cassandra binding with maintained one.

kw217:
@tupshin's Cassandra binding is no longer maintained (and does not currently build!), but there is a maintained fork. Update the README to point to the new one.


luciusmagn:
hmm, could we perhaps keep a mention of the original author in the README as well?

kw217:
Do you mean listing it as a separate project below the maintained one? I don't think that makes sense - the contribution guidelines say "If you want to add an entry to the README.md please consider this: is the entry valuable to people trying to get things done in Rust?" and that crate has been [broken since January](https://github.com/tupshin/cassandra-rs/issues) :-(

The maintained crate [does credit the original crate](https://github.com/Metaswitch/cassandra-rs#history) in its README, and it's a github fork of the original.

luciusmagn:
Very well, then. I guess that if it's credited and the original is broken, and the fork is in active development, then I believe it can be merged

kw217:
Thanks!

PR #370, opened 2017-11-02
add actix-web and sockjs links

fafhrd91:


luciusmagn:
A'ight, thanks for the addition

PR #377, opened 2017-11-21
Added reference to Relm (Gtk lib)

ivanderk:
Relm is an asynchronous, GTK+-based, GUI library inspired by Elm


luciusmagn:
I knew about Relm for a while know, somehow I thought it's in the list already. Mergin'

PR #378, opened 2017-11-21
add quadrupleslap/leven

quadrupleslap:
It's an application written in Rust, and it's awesome.


luciusmagn:
Thanks you, it indeed does look awesome

PR #380, opened 2017-11-30
fix redirects in README

luciusmagn:
let's see if I caught everything


luciusmagn:
hah, looks like I did!

kud1ing:
Thanks. Now only the link labels need to bee adjusted.

PR #381, opened 2017-12-03
Add extra games

Luke-Nukem:
Add extra games:
* magog
* rust-doom
* rust-quake

rust-doom and rust-quake may be better placed under a new heading? Perhaps "Game Renderer"?


luciusmagn:
Ah, nice. I am not sure about the cathegory, maybe when there is more of them. Fragmenting isn't good.

Could you fix the link for the travis badge image? It is supposed to be 'api.travis-ci.org...'. When you do so, the travis build should succeed and I can merge

Luke-Nukem:
Got it 😃

luciusmagn:
Awesome, thanks for the additions!

PR #387, opened 2017-12-08
add link to zbox

burmecia:
Zbox is a zero-details, privacy-focused embeddable file system. Please add this link, thanks.


luciusmagn:
Looks interesting indeed. Thank you for the addition

PR #399, opened 2018-01-10
add some awesome-macro

woodgear:
i think it should have a category for macro
cute and hado are the most useful macro in my work and also very awesome


luciusmagn:
Fine addition, thanks

PR #403, opened 2018-01-20
Add Perl libs to FFI section

ccakes:


luciusmagn:
Thank you for the addition! Build error is unrelated.

PR #405, opened 2018-01-23
Add eidolon to list

nicohman:
Hi guys. If my project isn't too small, I'd like to add my game manager/launcher written in Rust. Thanks for considering it!


luciusmagn:
Your project looks mighty fine to me. It doesn't matter if your project is small if it has a use. There is some beauty in frugality, after all. Thank you for your contribution

PR #407, opened 2018-01-26
Updated README.md to reflect the changes on ivanceras/curtain to ivanceras/diwata

pconcepcion:
https://github.com/ivanceras/curtain is deprecated and has been moved to https://github.com/ivanceras/diwata


luciusmagn:
Any chance we could keep both? Someone might still find a use for the old tool

pconcepcion:
We can have both, mabye putting a note that one is deprecated in favor of the other, but as I understood it's the same tool, just renamed, and the old one I understand it's deprecated and will not be updated, so I'm not sure why someone would be interested on the old one.

luciusmagn:
Hmm, it seems weird to abandon a repository, then continue it under a different name, but if that's indeed the case, I can accept these changes. Thank you for your contribution!

PR #413, opened 2018-02-13
Add just to development tools

casey:
Add github.com/casey/just to development tools.

It's a command runner with a nice file format for saving project-specific commands. I didn't include a link to the crate, since as a binary that doesn't seem super useful.

I'm not sure what the best short description is "A handy command runner for project-specific tasks" seems good, but I'm certainly open to suggestions.

Thanks!


luciusmagn:
Looks good. Thank you for your addition. Travis fail is unrelated

PR #417, opened 2018-02-19
Add enum-map

xfix:


luciusmagn:
Good job, the library looks pretty useful, I like it. Thank you for your contribution!

###### but I still think C isn't an inherently bad language :p

xfix:
> but I still think C isn't an inherently bad language :p

Did I say that somewhere? I have some opinions about C, but I tried to not show those in documentation, as it isn't really a place for political commentary. Besides, C doesn't really has something like this builtin into language (it's called an array), not type safe, but is anything in C ever type safe ;).

The library itself was modelled after [`EnumMap` in Java](https://docs.oracle.com/javase/7/docs/api/java/util/EnumMap.html), but of course with an interface working more like `HashMap` in Rust, because it's Rust, not Java.

luciusmagn:
I am the guy who argued with you about the design of C on Discord a few days ago - `M(Arc<Mutex<HashMap<A,Vec<m>>>>)`, haha. 😄 

Nothing to do with the library, it provides great functionality.

PR #418, opened 2018-02-20
unify and expand command-line section

memoryruins:


luciusmagn:
Hmm, good catch that TUI stuff was fragmented into both GUI and CLI sections. Thank you, good job

PR #420, opened 2018-03-02
jhspetersson/fselect

jhspetersson:
CLI tool to search files with SQL-like queries


luciusmagn:
Lookin good. Thank you

PR #422, opened 2018-03-06
Add llogiq/mutagen to testing

llogiq:
The version on cargo is still rough around the edges, but that will change soon, now we have a cargo-based mutation runner.


luciusmagn:
Ah, looks. Always been a fan of your stuff, llogiq. Thanks for your contribution.

Travis fail is unrelated

PR #424, opened 2018-03-08
Add FGRibreau/spinners.

memoryruins:
(and change rexpect's dash to match the length of the others)


luciusmagn:
Ah yes, thanks for fixing it. spinners look neat, thank you for the addition!

PR #426, opened 2018-03-08
Add m4b/goblin.

memoryruins:
Cross-platform, zero-copy, and endian-aware binary parsing.


luciusmagn:
Ah, I see you are on the hunt again. Looks pretty neat. Merging!

PR #428, opened 2018-03-13
Add JohnBSmith/moss to Scripting

JohnBSmith:


luciusmagn:
Your language looks very nice, I am a fellow connoisseur of scripting languages, but yours is better. Keep up the good work

PR #437, opened 2018-03-21
Add Tcod-rs under Game development

Oscuro87:
Tcod-rs: Bindings for the libtcod (c++/python, mainly) roguelike library. Added under Game development section.


luciusmagn:
Ah, all is good now. Thank you for the fix as well!

PR #439, opened 2018-03-24
Added Sozu : A HTTP Reverse Proxy

mothsART:


luciusmagn:
The link for the travis badge needs to start with 'api.travis-ci.org' instead of just 'travis-ci.org'. If you fix that, the travis build will succeed

PR #442, opened 2018-03-27
Remove Iron and handlebars-iron

equal-l2:
Iron is not active anymore.
See iron/iron#576.


luciusmagn:
Our current policy is not removing any entries from the list unless absolutely necessary. Iron might still be useful to some and remains a very powerful framework. Thank you for your concern about the freshness of the list, but I am closing this for now

rask:
Just a heads up: Iron is seemingly being actively maintained again.

PR #443, opened 2018-03-27
Fix badge for frank_jwt

equal-l2:
The badge is linked to Iron's.


luciusmagn:
Thank you for the fix, looks good!

Issue #404, opened 2018-01-21
Industrial automation category?


flosse:
I'd like to add the [tokio-modbus](https://github.com/slowtec/tokio-modbus)  library but I'm missing an appropriate category. Would you accept `Industrial automation` or `Automation engineering`?

luciusmagn:
Industrial Automation sounds good, although probably both would be fine. Sorry I missed this issue for so long

flosse:
Thanks for your feedback. I just crated #444 :)

PR #444, opened 2018-03-28
add tokio-modbus library

flosse:


luciusmagn:
The travis badge link needs fixing. Should be api.travis-ci.org etc. instead of just travis-ci.org. Otherwise, looks good.

Speaking of good, turning issue #404 into PR #444 is pretty funny. What are the chances of so many 4s? Not to mention the meaning of 404...

flosse:
> Speaking of good, turning issue #404 into PR #444 is pretty funny. What are the chances of so many 4s? Not to mention the meaning of 404...

yeah, that's indeed funny :)

luciusmagn:
Ah, I forgot to actually merge it! My bad. Thank you for your contribution

PR #448, opened 2018-04-01
Remove a couple duplicates

polyzen:
Parity tools already listed under Cryptocurrencies section.


luciusmagn:
While the usual policy is not to remove any entries, these ones are the exact same thing 15 lines apart. Thanks for pointing it out and providing the fix, I am merging

PR #450, opened 2018-04-12
add LSP clients for Vim/Neovim and Kakoune

ul:
Both clients could be used with Rust Language Server, and both are implemented in Rust.


luciusmagn:
Looks good. mergin'. Thank you for your contribution

PR #454, opened 2018-04-16
add coap crate

Covertness:


Covertness:
It seems some other links broken that not included in this pr.

PR #462, opened 2018-05-07
crates extension added for vscode

serayuzgur:
**crates** is an extension for crates.io dependencies. Aims to help developers to manage dependencies while using _Cargo.toml_. It is only helpful if you are using dependencies from _crates.io_.


luciusmagn:
Looks useful. Thank you for your contribution! The links are fine now, Travis fail is completely unrelated

PR #468, opened 2018-05-29
Add yoshuawuyts/human-panic

claui:
Dear maintainer(s),

I just [got word of this library](https://twitter.com/spacekookie/status/1001139793336717312) and feel it would be an extremely useful addition for the _Library_ » _Command line_ section. Also, [human-centered design](https://en.wikipedia.org/wiki/Human-centered_design) in IT is important (and underappreciated) enough to warrant a new subsection here, which this PR adds as well.

Pinging @yoshuawuyts and @spacekookie, the authors of the library, for visibility (and a chance to consent/dissent).


spacekookie:
Oh it would be super cool if `human-panic` could be included in this list :)

luciusmagn:
How come I haven't heard of this earlier!
The fact that it is not in the list already is a sin in and of itself! Thank you for your contribution.

PR #469, opened 2018-05-29
Add YaSerDe

MarcAntoine-Arnaud:


luciusmagn:
Looks good, thanks for your contribution. Is there a `serde` integration for your project? That would be pretty cool

MarcAntoine-Arnaud:
@luciusmagn I create this library due after discussing with SerDe guys.
The conclusion was: we don't want to add some specific attributes to manage XML features (namespace & attributes).
So I create a kind of a similar project to support full XML features, which can be compatible with SerDe to serialize models into over formats like json.

luciusmagn:
Ah, I see, makes sense. I was mainly talking about the handy Serialization/Deserialization derives. I have to admit that you do a good job

PR #470, opened 2018-05-30
adding OPC UA rust library

biancode:


luciusmagn:
You can never have enough industrial automation :man_shrugging:

PR #471, opened 2018-05-30
adding IIoT Modbus library

biancode:


luciusmagn:
Nice one, thanks for expanding the category! It used to be pretty empty...

PR #472, opened 2018-06-01
Add runiq binary crate

whitfin:
Hi! This will add the runiq crate.

I'm not sure if ordering is alphabetical by project or by author?


luciusmagn:
Nice little utility you got there. I like those.
Usually by author, if author is present, but this list drags a lot of historical baggage with it, so it might not be always consistent. Your entry would have the same position in both cases, so you needn't worry.

PR #481, opened 2018-06-22
Change redox os link to the new gitlab repository

AntiSC2:
Redox OS has switched over to gitlab and their Github is just a mirror of that now.


luciusmagn:
Thank you, but this already been fixed in a different commit

PR #483, opened 2018-06-24
Add atom/xray

VPashkov:
An experimental next-generation text editor.


luciusmagn:
There can never be enough good editors written in Rust, merging!

PR #490, opened 2018-07-09
Add slingr and add a video category

yuvadm:


luciusmagn:
Interesting, thank you for the addition

PR #491, opened 2018-07-11
Change redox-os link

liamnaddell:
the github is just a mirror so pointing people there seems wierd.


luciusmagn:
Good point, thanks for fixing it

PR #495, opened 2018-07-19
Add Remacs to "text editors" section.

mpereira:


luciusmagn:
Awesome, like I say, there can never be enough text editors written in Rust. Thanks for the addition!

PR #502, opened 2018-07-26
Add orion

brycx:
Full disclosure: I'm the author of the library.

As the description mentions, this library aims to provide crypto that is easy to use. This is done by providing "one-shot" style-like functions that are beginner friendly. It is still in an early stage, meaning this is considered an experimental library. I considered some of the other libraries that were listed under `Cryptography` and thought I'd give it a shot.


brycx:
The Travis build timed out, anything I can do here?

luciusmagn:
Nope, not your fault. It just happens sometimes. I will check out the PR when I get home later today.

I have already heard about `orion` and I think it would make a fine addition to the collection

brycx:
Thanks @luciusmagn.

PR #504, opened 2018-07-30
Add security section

kpcyrd:
Adds a section for security tools and 3 projects.


luciusmagn:
sounds good, thanks for the contribution!

PR #505, opened 2018-07-31
add a link to the duct crate

oconnor663:
`duct` is somewhat redundant with the `subprocess` crate, and I don't know what the process is for including redundant things, but I figured I'd submit a PR :)


oconnor663:
Also the contributing guidelines suggest alphabetical ordering, but I don't know if it's supposed to be by crate name or author, and the current file doesn't seem super alphabetical :-D

luciusmagn:
Yeah, that's kind of an historical issue. Redundant things are fine, we love alternatives. Thanks for the contribution

PR #506, opened 2018-08-01
CommonMark HTTPS support

brycx:
For the current Travis build to pass, this needs an update 😄


luciusmagn:
There is always something breaking :D thanks for the tiny fix

PR #508, opened 2018-08-11
Add rust-term-grid library to TUI section

ArniDagur:
This is the library [`exa`](https://github.com/ogham/exa) and Rust coreutils [`ls`](https://github.com/uutils/coreutils/blob/master/src/ls/ls.rs) are built on. It's pretty cool in my opinion.


luciusmagn:
Very nice addition, thank you for your contribution

PR #509, opened 2018-08-13
Added Kiss3D and Unrust to "game engines"

paulirotta:
Both may also be in WebGL and WebAssembly sections in a future re-org


luciusmagn:
Thanks for the addition!

PR #510, opened 2018-08-13
Add "cloak" to Application utilities

evansmurithi:


luciusmagn:
Nice project, thanks for adding it

PR #511, opened 2018-08-17
fix 301 and 404 links

swalladge:
Should make Travis happier. :)


swalladge:
Ok, it seems nothing will make Travis happy, but at least all the links are now working to my knowledge.

luciusmagn:
Aye, Travis be a harsh mistress :D

PR #514, opened 2018-08-22
servo/cocoa-rs has been deprecated

travispaul:
servo/cocoa-rs has been deprecated and moved into servo/core-foundation-rs


luciusmagn:
Alright, sounds good.

PR #516, opened 2018-08-24
add fblog and fw

icepuma:


luciusmagn:
Nice additions. I especially like the idea of `fw`.

PR #522, opened 2018-09-25
+ coinbase-pro-rs

inv2004:


inv2004:
A bit unclear - checked the link manually and they looks ok.

luciusmagn:
The point is that they are redirects (see the 301 HTTP Status, which means Moved Permanently), and when you keep these instead of the final destination, you are running the risk of the original link suddenly stopping working, especially since it's a 301 and not 302 or 307 (both of which are used for temporary redirects).

You don't need to worry much about the build failing, if it would've been something you've caused, I would've told you.

The only thing that needs changing in your PR is that the link to the SVG should be `api.travis-ci.org` as opposed to just `travis-ci.org`, but I'll just fix that myself, as I am already doing some cleanup work.

inv2004:
Thank you.

PR #524, opened 2018-10-02
Fixed broken rusters link, and updated redirections.

vsupalov:
So the TravisCI icon is green again.


luciusmagn:
Alright, sounds good. Thank you for your contribution!

PR #525, opened 2018-10-04
Fix markdown by removing a space in a link

carols10cents:
There's a space between the ] and the ( of this link, so it doesn't render as a link on GitHub. This should fix it! ❤️


luciusmagn:
Nice catch! Thanks for noticing it.

PR #527, opened 2018-10-06
Add Pop!_OS-developed crates

mmstick:


mmstick:
Most of what we write today is in Rust. It's a retiremen requirement for new Pop projects to be in Rust. The web team works with Elixir, though they've dabbled in Rust a bit for personal projects.

luciusmagn:
That's amazing, I am glad to see a Rust-powered distribution.

PR #530, opened 2018-10-08
Add a few more Pop!_OS crates / app

mmstick:


luciusmagn:
Thanks for the additions, if Pop!_OS/System76 has more Rust goodies, don't be afraid to send 'em this way!

PR #533, opened 2018-10-11
Add easy_reader

ps1dr3x:


luciusmagn:
Alright, this is pretty neat, I was actually looking for something like that myself a while ago.

Thank you for your addition to the list.

(Also one small nitpick, the word **fastly** doesn't actually exist, your description should say _... Easily and fast._ :) )

ps1dr3x:
Oops! Thanks for the correction, I was misled by the word "slowly" :)

PR #534, opened 2018-10-12
Add parallel-getter & progress-streams from Pop repos

mmstick:
Adds two new crates:

- parallel-getter
- progress-streams

`progress-streams` is a direct dependency of `parallel-getter`, to provide the callback functionality. We needed to develop the `parallel-getter` crate because downloading `deb` archives from an APT repo with a single GET request is really slow (500KiB even if you have a 50MiB connection). The `parallel-getter` crate was needed for our `debrep` utility to speed up the repo sync and fully utilize all of our available bandwidth. Mirrors are also supported to speed up downloads. This and the url-crawler may eventually be the basis for an `apt` replacement, should the Pop project get around to replacing it.


luciusmagn:
Nice projects, thank you for adding them to the list!

PR #535, opened 2018-10-14
Add CMakeRust

mlfaw:


luciusmagn:
Combining Rust with CMake, huh, strange, but thanks for adding it nevertheless!

PR #538, opened 2018-10-22
Add ddh utility

darakian:
This PR adds my ddh utility and fixes a bit of alphabetic ordering.


luciusmagn:
Nice utility with a fun name, thanks for the fix and the addition!

PR #546, opened 2018-11-03
Add cardano project

nurazhar:
Add cardano project


luciusmagn:
Well, thank you for adding these to the list!

PR #548, opened 2018-11-04
Add s3-concat to utilities

whitfin:


luciusmagn:
Nice, thank you for your contribution.

PR #551, opened 2018-11-08
Add joinery, a string join library

Lucretiel:


luciusmagn:
Hey, that's pretty good! I love small handy libraries like this one. Thank you for your contribution to the list.

PR #554, opened 2018-11-16
Add har-rs

mandrean:
This PR adds the `har-rs` lib and a corresponding new section for HAR/HTTP Archive Format files used by Charles Proxy, Google Chrome etc.


luciusmagn:
Looks good, thank you for adding it to the list!

PR #563, opened 2018-11-30
Proposal for addition of Plume and Firecracker

edomaur:
Plume is a blog agregator
Firecracker is an Amazon project for microvms


kud1ing:
Thanks.

PR #564, opened 2018-12-03
Add ChainX in Cyptocurrencies

gguoss:


luciusmagn:
Nice, thank you for your contribution

PR #566, opened 2018-12-07
Add NearProtocol to crypto projects

nearmax:
http://nearprotocol.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ See also [Friends of Rust](https://www.rust-lang.org/friends.html)(organizations
 
 ### Cryptocurrencies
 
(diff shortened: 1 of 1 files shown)


luciusmagn:
Nice project you guys got there, thank you for adding it to the list.

On a list-unrelated note and in layman-ish (as I am not really an expert in the field) terms, how is NEAR different from a traditional blockchain?

nearmax:
Users can "mine" on a low-end mobile phone with bad network connection, while at the same time have 100k tps throughput (compare to Visa or MasterCard that have 60k tps) and several seconds finality. This is enabled by:
* Using BFT consensus algorithms instead of proof of work. Proof of work requires heavy computations which would not be feasible on a mobile phone. Also, it has a low energy footprint. Compare to Bitcoin that uses proof of work -- Bitcoin farms consume more energy than Denmark.
* Sharding algorithms. Ethereum, EOS, and other platforms either require every device to be responsible for the entire network or use centralized cloud servers. Former, would not allow mining on a mobile phone, while latter is bad because of the centralization.

PR #568, opened 2018-12-07
added ruplacer entry

Dowwie:


luciusmagn:
This looks neat, I always had a soft-spot for text processing utilities. Thank you for adding it to the list.

PR #570, opened 2018-12-14
Add github.com/Peltoche/lsd to the System tool section

Peltoche:
lsd is a pretty ls based on nerd-fonts.


luciusmagn:
Hah, what nice and funny project. Thank you for adding it to the list

PR #571, opened 2018-12-15
Move yaserde from yaml to xml

mysteriouspants:
I'm reasonably certain that yaserde belongs under xml, rather than yaml.


luciusmagn:
Ah yes, you are absolutely correct. Thank you for your fix

PR #573, opened 2018-12-17
Add todo_r to text processing

lavifb:


luciusmagn:
This is a pretty neat idea, I like it. I'd be interested in adding support for more languages, could you give me any pointers?

PR #578, opened 2018-12-20
cargo-find is deprecated

pickfire:


luciusmagn:
Howdy, yes, cargo-find is deprecated, but that doesn't mean it should be removed, as we currently have a policy of not really removing stuff on account of the fact that even deprecated projects might have a value to someone if only educational.

Would just adding a note that it is deprecated suit you better?

pickfire:
Sure, any suggestion for deprecated? cargo shield?

luciusmagn:
Ah, apologies, missed the message. I guess just adding [DEPRECATED] or something might be fine. We don't currently have a format for this AFAIK.

Sorry for the delay,

luciusmagn:
Neat, that's pretty elegant. Thank you

PR #585, opened 2019-01-03
Adding aymanmadkour/glock

aymanmadkour:
Adding crate aymanmadkour/glock, which is a granular locking library.


aymanmadkour:
Thank You!

PR #586, opened 2019-01-04
Add whitfin/bytelines

whitfin:
Adds a small crate for processing lines of text as bytes rather than `String` due to faster throughput.


luciusmagn:
Looks pretty handy! Thank you for your contribution to the list.

PR #589, opened 2019-01-05
Replacing link to imageproc library

swarnimarun:
Imageproc library is now a library by PistonDevelopers based on Image library the one from the old link died years ago.


luciusmagn:
Ah, thanks for the quick fix!

PR #590, opened 2019-01-05
Add automotive section

marcelbuesing:
Add section that describes libraries that cover signal buses or standards used in automotive.


luciusmagn:
Oh my, that's a handful! Thank you for adding them all to the list!

PR #591, opened 2019-01-07
wrong place of `WebSocket`

zyctree:


luciusmagn:
Neat, thanks for the fix!

PR #592, opened 2019-01-10
Simple component-based GUI Toolkit for developing beautiful and user-…

lifeart:
…friendly interfaces.


luciusmagn:
There can never be enough enough GUI libraries :)

Thanks for the contribution!

PR #593, opened 2019-01-11
Some FreeBSD specific rust libraries

No9:


luciusmagn:
Very nice, I am very glad to see some BSD cool stuff here. It is especially great that there are bindings for capsicum now, which is an amazing thing.

PR #596, opened 2019-01-13
Add services category

saschagrunert:
Hey 👋 , how about adding a service category to collect all the available web services for the Rust ecosystem?


luciusmagn:
A swell idea. I like it. I had no idea deps.rs existed until now (shame on me, I know), which already shows that there is merit in having this section!

Thanks for adding it

PR #597, opened 2019-01-14
Add cargo-expand

theduke:


luciusmagn:
Nice, thank you for adding this. I actually used this only a week or two ago, it's extremely handy indeed.

Thank you for adding it.

PR #602, opened 2019-01-20
Move cargo-expand under Cargo section

dtolnay:
I believe this was mistakenly added under Webpack in #597.


luciusmagn:
Ah, yes. You are correct. Thanks for your fix!

PR #604, opened 2019-01-25
Added Rust extension for Visual Studio 2017

skuzins:


luciusmagn:
Ah, finally there is one! Shame I don't use VS anymore...

Glad you added it to the list!

PR #605, opened 2019-01-29
README: add Noria to database

Ivshti:
adding Noria, as seen here: https://www.youtube.com/watch?v=s19G6n0UjsM


luciusmagn:
Nice, it even comes with a video. Very fancy.

There seems to be an awful lot of Rust databases nowadays, it's pretty nice.

Thank you for adding it to the list!

PR #608, opened 2019-02-07
Include "Last commit" badge

p-kraszewski:
For github.com based repositories, add "Last commit" badge, to visually indicate if package is still maintained.

I've added a script (using gnu `sed` allowning `-i` inplace editing) to automatically clean and fill necessary entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,60 +95,60 @@ If you want to contribute, please read [this](CONTRIBUTING.md).
 
 See also [Rust — Production](https://www.rust-lang.org/production)(organizations running Rust in production).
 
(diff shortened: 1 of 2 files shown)


palfrey:
I'm not entirely convinced of the usefulness of this. For starters, you get some repos where they haven't been updated for a year, but that's fine because they still do exactly the task they're meant for. Also, the README is already a horrendous load time with all the build badges, and this would effectively double that. The latter item needs thinking about fully at some point, as we maybe need to figure out how to deal with that better, but until we can resolve this I don't think this is a good idea.

@luciusmagn Any thoughts?

firedupmike:
Exactly, with all the build badges, the README painting time on the browser is huge. Adding last commit badges would only add to the loading time

PR #609, opened 2019-02-08
Adding PickleDB

seladb:


luciusmagn:
There seems to be a huge influx of databases lately, I wonder why. Nevertheless, another great addition to our collection!

PR #611, opened 2019-02-12
Fix CI: remove absent project and change links for renamed projects

AnderEnder:
* remove absent `pop-os/pop-recovery`
* update link `rust-lua`
* update link `human-panic`
* update link `libui-rs`
* update link `rustlings`
* update link `Qt`


luciusmagn:
Thank you for the fixes!

PR #617, opened 2019-02-19
Added Sofa HTTP Client for CouchDB

sntdevco:
More choices for rust developers using CouchDB


luciusmagn:
That's Sofa king nice, thank you for adding it to the list! :^)

PR #618, opened 2019-02-20
Added MAY stackful coroutine library for Asynchronous

Xudong-Huang:
Added MAY stackful coroutine library for Asynchronous


luciusmagn:
Oh, very nice, thank you for adding it to the list!

PR #619, opened 2019-02-20
Add Bitcoin SV

nurazhar:


luciusmagn:
Bitcoin SV looks quite interesting, glad there is Rust support for it now. Thank you for your contribution to the list!

nurazhar:
Yup a famous anonymous just buidl instagram that stores image in the blockchain (immutable)

https://bitstagram.bitdb.network/

PR #620, opened 2019-02-20
fix watchexec build badge

LesnyRumcajs:


luciusmagn:
thanks for the quick fix!

PR #621, opened 2019-02-22
Removed extra forward slash from src attribute.

brudnak:
The extra forward slash in the the img tag's src attribute was creating
a broken link.


luciusmagn:
Nice catch, thank you for fixing it!

PR #622, opened 2019-02-26
Add archery

orium:


luciusmagn:
A very handy library, I like the concept. Thank you for adding it to the list and for the small fixes as well!

PR #623, opened 2019-02-26
Add ff (find-files)

vishaltelangre:
https://github.com/vishaltelangre/ff


luciusmagn:
Nice little utility, thanks for adding it!

PR #624, opened 2019-03-02
Add network simulation library

povilasb:
`netsim` is a crate for simulating networks for the sake of testing network-oriented Rust code. You can use it to run Rust functions in network-isolated containers, and assemble virtual networks for these functions to communicate over.


kud1ing:
Thanks.

PR #625, opened 2019-03-02
add cargo-deps, remove cargo-dot

m-cat:
+ Added cargo-deps.
+ Removed cargo-dot. It is unmaintained and the repo points to cargo-graph.
+ Add note that cargo-graph is unmaintained. I would remove it, but it still has a big presence in the community (many GitHub stars).


kud1ing:
Thanks.

PR #626, opened 2019-03-03
Rewrite cleanup script in rust

holykol:
Get rid of python and dummy.rs :tada: 

New script usage:
`rustc cleanup.rs && ./cleanup`


kud1ing:
Very nice, thank you.

kud1ing:
Would you be interested in expanding the cleanup script?

* It would be nice if the entries were sorted automatically.
* Also it woud be nice if we could ensure that the entries have the correct format, as described in `CONTRIBUTING.md`.

PR #645, opened 2019-04-07
Update amethyst's CI badge

LucioFranco:
We've recently move to jenkins ci from travis thought we might as well update the badge here too!


kud1ing:
Thank you.

PR #647, opened 2019-04-09
Adding Joystream Rust Substrate runtime

bedeho:
Here is more info on the entire project
https://github.com/Joystream/joystream


kud1ing:
Thank you.

PR #648, opened 2019-04-10
Fix repo link for sd

sntdevco:
Why are we linking to the crate? Replace link for repo instead


sntdevco:
Closing as updated contributions require a link to the crate (don't know why though)

PR #650, opened 2019-04-11
Replace cbor with serde_cbor

sanxiyn:
As shown in [README](https://github.com/BurntSushi/rust-cbor), cbor is unmaintained and serde_cbor should be used instead.


luciusmagn:
Sounds good, thank you for the quick fix!

PR #651, opened 2019-04-13
Offer the sendmmsg crate for low-level networking

Gymmasssorla:


luciusmagn:
Low-level networking is cool!

Thank you for adding your crate to the list

PR #654, opened 2019-04-14
Adding consoletimer utility

arthrp:
Consoletimer is a simple timer for a terminal (<100 LOC), written using ncurses.


luciusmagn:
Ah, tiny utilities like this are great.

Thank you for adding yours to the list.

PR #655, opened 2019-04-16
vlopes11/futures-jsonrpc Futures implementation for JSON-RPC

vlopes11:


luciusmagn:
Pretty neat and well documented. This is sure to come in handy, thank you for adding it to the list!

PR #656, opened 2019-04-20
Add woboq/qmetaobject-rs

Ace4896:
Crate that integrates Rust and QML via QMetaObjects.

Side note: I noticed that Qml-Rust is at the bottom of the Qt section, unlike the other crates whose names start with q (which are at the top of the section). Should I move this up as well?


luciusmagn:
I think it's fine, there's some weirdness in ordering anyway.

PR #659, opened 2019-04-28
Add neachdainn/nng-rs

neachdainn:
NNG is the successor to Nanomsg and, given that [Nanomsg is actively recommending switching to NNG][1], it seems like it would be a good idea to have it on this list. Thanks!

[1]: https://github.com/nanomsg/nanomsg#support


luciusmagn:
Indeed it would, thank you for adding it!

PR #661, opened 2019-05-02
Add schulke-214/rsnake

schulke-214:
A basic implementation of the classical snake game.


luciusmagn:
Ah, very nice! Good old snake!

PR #662, opened 2019-05-02
[chore] update to new github action

icepuma:
I updated the github action link


luciusmagn:
Thanks for the quick fix!

PR #663, opened 2019-05-06
Add dalance/procs

dalance:
procs is a replacement for ps written by Rust.


luciusmagn:
Awesome, I always like these cli utilities!

Thanks for adding it to the list!

dalance:
Thanks!

PR #664, opened 2019-05-07
Add cb to utilities

yaa110:


luciusmagn:
This is very handy, thanks for adding it!

PR #667, opened 2019-05-13
add kytan

attenuation:
kytan: High Performance Peer-to-Peer VPN


luciusmagn:
Yeah, we really don't want to allow redirects. The point of the `awesome_bot` is to check for broken links and redirects. That is because we can assume that a link that redirects is a link that will be eventually broken. If you kill the `allow redirect` commit, I can merge.

Thanks!

attenuation:
but if i kill that commit, the CI build task will failed

luciusmagn:
Yeah, that is the point, the build should fail so that we know what links are the issues and may change the links.

Thank you for your contribution to the list and the small fixes!

PR #670, opened 2019-05-20
Add nokia/memory-profiler to Profiling

YoshiTheChinchilla:
Add [nokia/memory-profiler](https://github.com/nokia/memory-profiler) to [Profiling](https://github.com/rust-unofficial/awesome-rust/blob/master/README.md#profiling)


luciusmagn:
Hey, that's pretty neat! Love to see nokia using Rust

PR #672, opened 2019-05-24
Add pwoolcoc/soup

bepvte:
[soup](https://gitlab.com/pwoolcoc/soup) is a very cool library that lets you query and manipulate html with an api similar to pythons beautifulsoup or jQuery


luciusmagn:
Nice soup

PR #674, opened 2019-05-25
add static file server see

wyhaya:


luciusmagn:
Good to have another static server! Thank you for your contribution

PR #677, opened 2019-05-29
Add the build status for ParityDB

jamesray1:


luciusmagn:
Thanks for the quick fix!

PR #681, opened 2019-06-06
Added WASM-PDF to Graphics/PDF section

jussiniinikoski:
WASM-PDF was created with Rust and wasm-bindgen: https://github.com/jussiniinikoski/wasm-pdf


luciusmagn:
Nice one, thank you for adding it to the list

PR #682, opened 2019-06-06
Added Holochain

f13end:
https://github.com/holochain
holochain.org


luciusmagn:
An evolution of block chain, interesting. Thanks for adding it to the list!

PR #683, opened 2019-06-09
Add schulke-214/connect-four

schulke-214:


luciusmagn:
Haha, nice, thanks for adding it!

PR #685, opened 2019-06-16
Added deno as Application

kamaruzzaman:
A Node.js alternative application written in Rust, TypeScript.


luciusmagn:
Hey, this is really neat. How does it compare performance-wise?

Nevertheless, thanks a lot for adding this one!

kamaruzzaman:
> Hey, this is really neat. How does it compare performance-wise?
> 
> Nevertheless, thanks a lot for adding this one!

I think deno is still under development. Its main goal is to give a secure alternative to Node.js which is not so secure.

PR #686, opened 2019-06-17
Add Nimiq to cryptocurrency projects

paberr:
I also sorted the entries for this section again alphabetically as it was quite a mess.


luciusmagn:
Alrighty, thanks a lot for your contribution to the list and the small ordering fix!

PR #688, opened 2019-06-23
add simd-json, simsearch

Ivshti:
Adding a simd-json and simsearch (name similarity not intended)


luciusmagn:
Haha, name similarity not intended yet still appreciated! Thank you for your contribution to the list!

PR #690, opened 2019-06-29
Add `sfackler/cargo-tree`

jakubadamw:


luciusmagn:
This is a very nice utility, I have used it in the past myself. Thanks for adding it, I had no idea it wasn't in the list already!

PR #692, opened 2019-07-07
Move WASM-PDF under PDF section

jussiniinikoski:
I missplaced it under Vulkan, sorry! Also added Travis ci badge.


luciusmagn:
Hah, that's fine. Thanks for the badge!

PR #693, opened 2019-07-10
Remove sendmmsg (404 not found)

Gymmasssorla:


luciusmagn:
Oh, you deleted it? What a shame, but nevertheless, thanks for the update!

Gymmasssorla:
If someone wants to use raw system calls `sendmmsg/recvmmsg`, he/she typically needs all the other stuff like `setsockopt/getsockopt`, `ioctl`, etc, so I decided that it's better to help [these guys](https://github.com/nix-rust/nix/pull/1017) than develop a wrapper detached from the real-life :)

luciusmagn:
Awesome, and we already got `nix` on the list, so all's good!

Cheers!

PR #695, opened 2019-07-11
Add plotlib data visualization by milliams

clumsy:
Here's a promising pure Rust plotting library


luciusmagn:
Damn, that looks promising indeed. Thanks a lot for adding it!

PR #697, opened 2019-07-15
battery crate

svartalf:
I'm not quite sure about category, since this crate on one hand wraps platform-specific APIs, but on the other hand works with all platforms listed.


luciusmagn:
Tough placement huh :D Well, I guess that's a fine place for now.

PR #699, opened 2019-07-22
OMG, smoltcp isn't here!?

Gymmasssorla:


luciusmagn:
Good point, thank you for your contribution

PR #700, opened 2019-07-22
Add zerocopy crate, "Unsafe" section

joshlf:


luciusmagn:
Neat, thanks for adding the crate, I am sure it will come in handy to many people. Kudos for adding the section, good idea, I might think of a few crates that could be added/moved there

PR #706, opened 2019-08-06
Audio: add Polaris -- music streaming application

ngirard:


luciusmagn:
Neat, a Rust alternative to MPD! Great to see that!

PR #710, opened 2019-08-18
Add Android / iOS example repo

i-schuetz:
An example of using a shared Rust lib for Android and iOS using rust-swig and cbindgen respectively.

https://github.com/i-schuetz/rust_android_ios


luciusmagn:
Neat, I am very looking forward to Rust mobile development!

PR #711, opened 2019-08-20
Add starship

bbigras:
Not 100% sure if "Prompt" means what I think. Starship is a terminal prompt:

image


luciusmagn:
Starship is splendid, thanks a lot!

bbigras:
Thanks!

PR #721, opened 2019-09-20
Move elastic_reqwest to a new repository

Stzx:


luciusmagn:
Thanks a lot for the fix!

PR #724, opened 2019-10-03
Add rx Pixel Art Editor to the list

swarnimarun:
I am not sure of where to place it in the list, but I guess Applications is as good a place as any.

https://github.com/cloudhead/rx


luciusmagn:

Glad to see pixel art editors being made in Rust. It's a real shame that Aseprite went commercial a while ago, so an OSS alternative was called for!

swarnimarun:
Haha yeah. There is also Slate but it's in Qt/QML + C++..

PR #725, opened 2019-10-04
Add fake-rs

theanht1:
This library is used for generating fake data like name, email, phone number


luciusmagn:
Neat! Thank you for adding it to the list!

PR #726, opened 2019-10-06
Added softprops/openapi to Web Programming/Misc

alamminsalo:


luciusmagn:
Awesome, glad to see some OpenAPI support in Rust!

Thank you for adding it to the list!

PR #727, opened 2019-10-07
include soydos in games section

benbrunton:
Soydos is a wasm implementation of Pusoy Dos (Filipino poker)


luciusmagn:
Hah, a neat little game, thanks for adding it. What does Filipino poker do differently from the US version?

PR #728, opened 2019-10-08
Add cryptography resources

defund:
Just wanted to show some love to pure Rust crypto <3, also plugging my own tool


luciusmagn:
And I have some love for your love!

PR #730, opened 2019-10-09
Add Rust for professionals

ifdiego:
Hey,

Including Rust for professionals link.
It's a link with Rust examples and comparisons with other programming languages like JavaScript, TypeScript, etc.


luciusmagn:
This is a neat resource, if a bit oversimplified. Nevertheless, I think that it will be valuable for many!

PR #731, opened 2019-10-12
Add Emplace

tversteeg:
Shameless plug.


luciusmagn:
Shame. :D

But I have seen your project a while ago and though it would be a great addition. Have you considered making a Reddit post promoting it as well? I think many would find it useful.

luciusmagn:
Also kudos for using RON file format!

tversteeg:
Thanks for the kind words!

PR #735, opened 2019-10-16
GUI: remove rise-ui, add OrbTk

Felk:
I removed rise-ui due to lack of development, and instead added OrbTk, which looks promising if one doesn't mind the dependency on SDL2


luciusmagn:
Thanks a lot for your contribution to the list, it looks like a pretty neat framework.

As for removing rise-ui, the current policy is that we don't remove projects unless proven broken or pronounced superseded by different project by their author as they can provide an educational value. This is a known and open issue.

Thank you!

PR #737, opened 2019-10-18
Fixed awesome_bot issues on travis

chrsmutti:
Fixed all the issues with URLs highlighted by awesome_bot.

- Whitelisted github redirects.
- Removed yarte (it's private on gitlab, or inactive).
- Update bzip2 homepage.


luciusmagn:
Sorry, we don't whitelist github as we need to fix the issues there. Unlike crates.io, sciter et al. its HTTP implementation isn't broken, so we don't get false positives. Ergo, the policy on github redirects is fixing them.

Thank you for the bzip2 fix, though!

PR #745, opened 2019-10-28
Add Rudr

ewbankkit:
Add [Rudr](https://github.com/oam-dev/rudr), a Kubernetes implementation of the [Open Application Model](https://openappmodel.io/) spec.
As one of the first significant Rust projects in the Kubernetes ecosystem, it has real reference value.

References https://github.com/oam-dev/rudr/issues/376.


luciusmagn:
Great reference value indeed. I may not know much about Kubernetes, but I see great value in Rust and Kubernetes coop.

Thanks a lot for adding this project!

PR #749, opened 2019-10-30
Feature/websocket core

dendisuhubdy:
Adding websocket Rust library for periodic message broadcast


luciusmagn:
I was actually looking for one such library a while ago, glad to see this one getting added!

PR #753, opened 2019-11-02
add QUIC (HTTP/3) implementations

IgorPerikov:


IgorPerikov:
all the issues according to those additions have been fixed

luciusmagn:
My my, QUIC is so new and we already have some implementations in Rust, very glad to hear that!

Thanks a lot for adding these projects! I appreciate your contribution

burdges:
At present it appears only quinn is doing the required dev work for general usability, like more general certificate scheme support, proper async integration, etc., not saying the others should not appear here, but that they're quite special purpose for HTTP in browsers.

PR #757, opened 2019-11-07
Add rspark library

reugn:
▁▂▆▇▁▄█▁ Sparklines for Rust apps.


luciusmagn:
Delightful,
I like these tiny utilities!

Thanks a lot for adding this crate!

PR #764, opened 2019-11-16
Add etcd-rs

ccc13:
Add [etcd-rs](https://github.com/ccc13/etcd-rs)


luciusmagn:
Was wondering, when would etcd show up.

Thank you for your adding this crate here'

PR #765, opened 2019-11-19
Add radiate 

pkalivas:
Added a link to a new crate I have recently released for genetic algorithms. Took some inspiration from this page and so wanted to be a apart of it also. The project is being developed pretty consistently so I think it would also benefit from the eyes of others, hopefully this list will give it some reach to those who can help build/make use of it!


pkalivas:
Hello, is there a reason this build has failed?

luciusmagn:
Yes, but the reason is unrelated.

Thank you for making a genetic algorithm crate, I love to see these and I am a great fan of the concept!

PR #767, opened 2019-11-22
Remove anevicon and finshir since they already exist in the category

gdimitris:
Noticed that anevicon and finshir projects already exist in the Security section so I removed them.


luciusmagn:
Unfiortunately, it appears that Gymnasssorla has since deleted their repositories, so both of these projects are no longer available. @palfrey has already issued a fix for that.

Thanks a lot for your vigilance though!

PR #768, opened 2019-11-23
Add the genetic programming library radiate

Martin1887:


luciusmagn:
Thank you but this library has already been added by its author, which takes precedence.

I appreciate your effort though and I am looking forward to more contributions from you!

PR #771, opened 2019-12-08
Added ripasso

alexanderkjall:
Added ripasso, a password manager, under the security section.


luciusmagn:
Thanks a lot for adding this!

Issue #763, opened 2019-11-15
global hotkey


geeksesi:
hello,
someon please add this :
[global hotkey](https://github.com/unwrittenfun/hotkey-rs)
@UnwrittenFun

Issue #742, opened 2019-10-25
Could you add Balista, distributed in memmory computation engine base on K8s in rust?


GrigorievNick:
https://github.com/andygrove/ballista

PR #773, opened 2019-12-13
Solve #742, #763

YoshiTheChinchilla:


luciusmagn:
Thanks a lot for adding these!

PR #774, opened 2019-12-15
Add Zwitterion

lastmjs:


luciusmagn:
I am really on the fence with this project, to be honest.

On one hand, the Rust link seems to be rather weak - the repository contains pretty much no Rust and the support for Rust appears to be very basic.

On the other hand, I don't really understand what is the server doing (so I can be mistaken about the support) and it already has >400 stars, which indicates that is starting to gain on importance.

Could you perhaps shed more light as to how it relates to Rust?

lastmjs:
Zwitterion is designed to be the simplest way to import code from any language into web applications. Rust is one of the supported languages in Zwitterion: https://github.com/lastmjs/zwitterion/blob/master/README.md#rust

For web app developers, the mainstream way to get other languages into your web application requires using a bundler like Webpack, Parcel, or Rollup. These bundlers can be complicated to deal with.

Zwitterion is essentially just a static file server, like a simple http server. As files are requested it will compile and prepare them for use in the browser. Any Rust files requested from the browser are compiled to WebAssembly and wrapped in an ES module for simple importing. IMO this is the simplest way to use Rust in the browser. Yes the support is basic, but it is my intention to improve support drastically over time.

Zwitterion literally let's you import Rust code into the browser and have it execute.

luciusmagn:
Alright, that does have some merit. @palfrey what do you think?

palfrey:
I think that's got enough merit to be added.  Although it would probably be good to replace the `*` at the end of the line with something like `(including Rust)` and add the build status badge.

luciusmagn:
I agree, that sounds reasonable.

> it would probably be good to replace the * at the end of the line with something like (including Rust) and add the build status badge

@lastmjs it would be neat to do that. Would you mind making these changes?

palfrey:
Given this hasn't had any changes in 3 months, I'm closing this PR. Please reopen if you're able to make the requested changes.

PR #776, opened 2019-12-20
Add a mention of the MIRAI static analyzer

huitseeker:


luciusmagn:
Thanks for mentioning it!

Sounds like a great Rust project so far!

PR #787, opened 2020-01-02
Includes EDN-RS and Hiccup

naomijub:
EDN-RS: is great for microservices written in rust that are in a Clojure environment, as Clojure's most used data notation is EDN. Parser is highly usable for our current projects and serializer will be implemented with Datomic client, in the near future.

Hiccup: Templating HTML, and also XML,  engine that prevents unclosed tags problems. It is inspired by the clean and simple syntax of Clojure's Hiccup Lib

### DIFF
--- a/README.md
+++ b/README.md
@@ -961,7 +961,7 @@ See also [About Rust’s Machine Learning Community](https://medium.com/@autumn_
 * [FlatBuffers](https://google.github.io/flatbuffers/)
   * [frol/flatc-rust](https://github.com/frol/flatc-rust) — FlatBuffers compiler (flatc) integration for Cargo build scripts [<img src="https://api.travis-ci.org/frol/flatc-rust.svg?branch=master">](https://travis-ci.org/frol/flatc-rust)
 * EDN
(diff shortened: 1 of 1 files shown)


palfrey:
Can you please also add the build badges for both repositories as I see they've got Github workflows?

naomijub:
4cbd389

palfrey:
Thanks for adding the build badges. However, the reason that the build is failing is that the badges are using the `https://travis-ci.org/` urls for the images, which redirects then to `https://api.travis.org/`. Could you correct these so it uses the end url directly as opposed to the redirecting one?

The relevant log is at https://travis-ci.org/rust-unofficial/awesome-rust/builds/644118937 and looks like this

[301] https://travis-ci.org/naomijub/edn-rs.svg?branch=master -> https://api.travis-ci.org/naomijub/edn-rs.svg?branch=master
[301] https://travis-ci.org/naomijub/hiccup.svg?branch=master -> https://api.travis-ci.org/naomijub/hiccup.svg?branch=master

Once the build is green we should be able to merge this.

naomijub:
Guess the current error is not from this PR anymore :(

palfrey:
> @palfrey Guess the current error is not from this PR anymore :(

Yup. That was #817 which just got merged, as something got rearranged there...

palfrey:
Thanks for your contribution!

PR #789, opened 2020-01-08
Many fixes and updates, mostly to fix the build

palfrey:
Primary goal was fixing the build, but this should also correct a bunch of broken build/repo links


palfrey:
Right, I'll stop messing around with things for the moment :) Ready to be reviewed/merged

luciusmagn:
Awesome, thanks a lot for your work! I have also checked out your rustified checker, see my response there

PR #791, opened 2020-01-09
Add dtolnay/cxx 

pravic:


palfrey:
Could you please add the Travis build badge here as well as it appears to have a working build there?

pravic:
@palfrey added!

PR #792, opened 2020-01-11
Rustified checker

palfrey:
(This builds on top of #789 so we've got a fixed README.md)

Given this is meant to be awesome Rust stuff, and that the Ruby checker appears to have various issues (e.g. when repeatedly checking the README it wants to do *all* of them every time), I wrote a Rust version of the checker. This doesn't do anything majorly different yet, but could be improved in the future to do things like cope with Github actions links not being available for checking for not logged in users, and possibly find the broken build links that I did a lot of manually in #789


luciusmagn:
Hey, this is awesome, long have I wished for a Rusty awesome-bot :D Is it fully compatible with the original? As in, will it properly flag everything wrong and inform about redirects?

palfrey:
> Hey, this is awesome, long have I wished for a Rusty awesome-bot :D Is it fully compatible with the original? As in, will it properly flag everything wrong and inform about redirects?

Mostly. So it detected actualy broken links, but I've just fixed it to detect redirects (and anything else other than 200's). One thing it doesn't pickup yet is urls in image tags, but I'll look at that tonight.

luciusmagn:
Awesome, I am looking forward to merging this very much.

What we need to detect are just redirects (with source and target url clearly visible) and 4XXs & 5XXs should also fail and print faulty URL, but further data is not required.

Posibility to whitelist sites that don't conform to HTTP spec (and may therefore result in false positives) would be also nice.

All in all, thank you for your work, I appreciate you doing this very much!

palfrey:
> What we need to detect are just redirects (with source and target url clearly visible) and 4XXs & 5XXs should also fail and print faulty URL, but further data is not required.

Done

> Posibility to whitelist sites that don't conform to HTTP spec (and may therefore result in false positives) would be also nice.

So far I've been able to make changes that have reduced those a lot. Adding a user-agent and suitable Accept header fixed a lot of them. The one annoying case is Github actions, which just aren't accessible to users that aren't logged in, and also don't have much useful data in their API regarding the workflow name, so they're really hard to check. There's a hack workaround for them which tries the main repo name instead and basically just hopes that if the repo exists the action does as well!

> All in all, thank you for your work, I appreciate you doing this very much!

Not a problem. This interested me, and it's taught me a bit more about async/await.

So, at this point all the failures should be "real" ones in the sense of they're actually broken. If you're happy with the current test output, I can go ahead and fix those in the README before this gets merged.

palfrey:
Whoops, one last item. And travis appears to have had a bit of a moment, so if you could re-run the PR build that'd be good thanks!

luciusmagn:
Hi and thank you,

this is great, I like the changes. I have but one final request, if it wouldn't bother you.

Would it be possible to change the error format to be less verbose?

Something like:

```
[XXX] https://url.com/site
[301]  https://original-url.com/old-site -> https://new-url.com/new-site
```

A less verbose output makes it easier to read and process with Posix filter tools like `awk`, `cut` et al.

However, I greatly appreciate the work you have done so far, so I'll gladly merge it as it is, if you disagree with my perspective. I don't want to seem nagging :sweat_smile:

palfrey:
> Would it be possible to change the error format to be less verbose?
> 
> Something like:
> 
> ```
> [XXX] https://url.com/site
> [301] https://original-url.com/old-site -> https://new-url.com/new-site
> ```
> 
> A less verbose output makes it easier to read and process with Posix filter tools like `awk`, `cut` et al.

I've redone the error work in line with that. Well, it still has more verbose errors when there's an outright failure to connect to something, but for anything with a status code it's inline with what you detailed.

> However, I greatly appreciate the work you have done so far, so I'll gladly merge it as it is, if you disagree with my perspective. I don't want to seem nagging 😅

No worries. Want to try and get this right, but OTOH I also regard this as a basis for later work on any other linting and other forms of automation to try and deal with some of the overheads of maintaining this list, so would be good to try and get this initial version in soon.

palfrey:
I realised that the markdown processor wasn't picking up links in HTML blocks (I don't think they should be used TBH, but they're there at the moment), so added in 6cb9433de0bbf8636ff15ead79d3297983075fd9 to fix that.

luciusmagn:
Exquisite, I took a look at the current version and I think it is good to go now.

I'll be merging this now, thanks a lot!

PR #800, opened 2020-01-13
Remove Anevicon

dmweis:
Seems that this project has been removed from GitHub. 
Both the owners account and project are gone


luciusmagn:
Yep, and so is Gymnassorla's other project! It is unfortunate that this code is lost from github, but what can you do.

@palfrey has already issued a fix for this.

Thanks a lot for your vigilance though!

PR #805, opened 2020-01-20
Detect "unknown" status Travis builds

palfrey:
Travis build badges that no longer refer to a build return 200, but have a Content-Disposition header that says `inline; filename="unknown.svg"` and so we can detect them with some changes to the checking tool. This builds on top of #804 so that we have an otherwise clean README. https://travis-ci.org/rust-unofficial/awesome-rust/builds/639610636 demos the failure case.


luciusmagn:
Cheers

PR #810, opened 2020-01-25
Add Rust Music Theory library, update Audio headers

ozankasikci:
[This](https://github.com/ozankasikci/rust-music-theory) library and potential future libraries won't fit under the Audio category, hence i felt the need to update the `Audio` headers to `Audio and Music`.


palfrey:
Thanks!

PR #812, opened 2020-01-26
Add 'tiny' IRC client

osa1:
(I'm the author of 'tiny`)


luciusmagn:
Hahah, I had no idea `tiny` wasn't on the list already!

I'm actually long time user and I like your client very much, although I wish there was a no dbus option.

Thanks a lot for adding it to the list!

osa1:
> I'm actually long time user and I like your client very much, although I wish there was a no dbus option.

Glad to hear it :-) Making dbus optional is planned as it's making static linking harder. It should not be too hard, be sure to watch the repo for updates :-)

PR #825, opened 2020-02-10
Add espanso

muhlinux:


palfrey:
Hmm, that's a fun one. Looks like the build is failing because of the Azure build redirect, but that there isn't a better static candidate URL. I'm going to poke at the linter now and see if I can fix that.

PR #828, opened 2020-02-12
Make sure Travis/Github build badges have a branch

palfrey:
IIRC Github does default sensibly to just master, but Travis will default to "last build, on any branch", which means you get marked as failing even if all you've got is a dodgy unmerged PR. This PR checks that both Github Actions and Travis build badges specify a branch, and fixes all the cases for both. I've also linked a few of the Github actions buttons back to their workflow pages.


palfrey:
@luciusmagn Any thoughts on this? This helps automate checking something I was mostly doing already, and make sure it remains fixed.

PR #838, opened 2020-02-21
Add teloxide

Hirrolot:


luciusmagn:
Hi, thanks a lot for your contribution to the list!

Unfortunately, I think that you added the project to the wrong category, since I doubt a Telegram bot would statically generate websites.

I think that putting it a few lines up, into the miscellaneous subsection would be more appropriate, thanks! It's really interesting how we don't have a bot section yet..

Hirrolot:
Oh, sorry, just misclicked.

Issue #859, opened 2020-03-26
429 errors from Github


palfrey:
We're seeing 429 errors when trying to request Github sites. This is because we're doing >60 requests/hour, which is the limit for unauthenticated requests. We should start optionally authenticating requests (and doing this in CI) so that we can get to the 5000 requests/hour allowed there, and probably also auto-rewrite all Github URLs to request the API instead so we can more easily use the authentication.

PR #860, opened 2020-03-30
Add zoxide under System Tools

ajeetdsouza:
zoxide is a CLI utility that tracks your directory usage, allowing you to quickly `cd` to anywhere on your filesystem using just part of the directory name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,7 @@ See also [A comparison of operating systems written in Rust](https://github.com/
 
 ### System tools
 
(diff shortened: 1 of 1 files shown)


ajeetdsouza:
I've double-checked the links and they seem to work; maybe a rerun will fix the build?

palfrey:
The earlier build failure was due to #859 which we need to look at at some point. Only started happening recently.

PR #867, opened 2020-04-03
Support optional use of Github secrets to workaround rate limits

### LINKED ISSUE #859: 429 errors from Github

palfrey:
This would fix #859 (429 errors from Github) but I don't have enough permissions to add a new secret value to Travis.

@luciusmagn Any thoughts? Do you have enough access for this?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -14,6 +14,7 @@ use regex::Regex;
 use scraper::{Html, Selector};
 use failure::{Fail, Error, format_err};
 use chrono::{Local, DateTime, Duration};
+use std::env;
 
 #[derive(Debug, Fail, Serialize, Deserialize)]
 enum CheckerError {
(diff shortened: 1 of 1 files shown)


palfrey:
@kud1ing Do you still have enough access to add a secret? Relatedly, I don't appear to be able to restart Travis builds anymore which I used to be able to :(

kud1ing:
Hi, no i am sorry. This is bound to the organization, which i've left.

palfrey:
@Manishearth Can you help here?

palfrey:
I'm ok with adding my own secrets to the travis.yml, but only if they're encrypted with the standard Travis encryption stuff. I can't do that however as I don't have enough access on travis, probably because I don't have enough access on this repository. If that can be fixed, I can fix the rest of things myself.

badboy:
We should be able to make @palfrey an admin on this repo. @Manishearth, you're good with that?

palfrey:
@badboy @Manishearth Any further thoughts on this?

Manishearth:
Yeah, this has been in my queue, i just have a ton of notifications to deal with.

You've been maintaining this pretty well, I'm okay with granting admin access. Please use it responsibly :smile:

palfrey:
Cool, thanks. I've added the secrets now (in the Travis UI, not the .travis.yml)

PR #895, opened 2020-05-13
Add mun-lang to Scripting section

daboross:
I'm not affiliated with mun, just stumbled across them.

It looks like they're still WIP / not committing to production stability. However, the language has had several releases, has active development, and is already a great example of a scripting language implemented in Rust.

I wasn't sure whether to list the github page or their website, https://mun-lang.org/. I went with the github page to match existing entries.


daboross:
Thanks!

PR #897, opened 2020-05-18
Add sscache support

palfrey:
This seems to get the build of the tool down from ~3m to ~1m15, and seems to work a lot better than the default Rust artifact caching (also has a limit to cache size which should result in it throwing away old data eventually, unlike stock cargo work)


palfrey:
Build failures are all the 429's and mostly a matter of the change in cache paths resulting in a full rebuild as opposed to being able to reuse what we get from master, so merging.

PR #915, opened 2020-06-25
Patch 2

aleshaleksey:
Add a parser (and roller) for various dice strings, which supports addition and subtraction of multiple dice groups with modifiers such as minimum value per dice, maximum value per dice, re-rolls for groups and explosive dice. It also supports generation of probability distributions and has very simple API (incomplete) for use in C/C++. 

In principle this library can be used to parse a string dice roll and then use the resulting "dice bag" for rolls. It should also be useable for making a dice bag using a "button" UI and then rolling it.

In practice it is currently set up to work better as a parser.

NB- I don't know whether this is a little bit trivial for `rust-awesome` but the dice parser on roll20 fails to impress. [Edit: That is not criticism of dice parsing done by various platforms. More to say that this is an attempt to improve on what exists ad provide functionality that some [read: I] would like to see.]


palfrey:
It's rather specialised, but the detail put into it and the decent test suite means I think it's mature enough to get added. I'm somewhat a fan of the roll20 dice bot myself, but I appreciate some have needs for more complicated rolls :)

PR #949, opened 2020-08-31
Updated the contribution guide with a new template

FlashSystems:
As offered I updated the contribution guide to use the repository link as a primary link and add an optional `crates.io` link. I hope that's ok.


palfrey:
Awesome, thanks

PR #955, opened 2020-09-07
Changed sort of list entry from library to application

replicadse:
I accidentally proposed a tool to the library instead of application section + corrected a spelling mistake.


palfrey:
Thanks! I didn't spot that...

PR #959, opened 2020-09-13
Added z5labs/pflag

Zaba505:
Hi!

I know this project isn't at `v1` yet but it is stable and usable. I haven't `v1`-ed it yet because right now it's more or less a direct translation and I'm looking for community feedback on if there could be a more Rustic API for it.

Thanks!


palfrey:
> I know this project isn't at `v1` yet but it is stable and usable. I haven't `v1`-ed it yet because right now it's more or less a direct translation and I'm looking for community feedback on if there could be a more Rustic API for it.

Happy to add at this point. I would suggest adding a basic usage section to both the front of the docs and the README for easier use.

PR #960, opened 2020-09-13
Add lethe

Kostassoid:
Please consider this humble offering to the list. The app went through multiple versions and improvements and I think is stable enough.


palfrey:
LGTM, thanks for the submission

Issue #963, opened 2020-09-18
Remove those badges from README.md


qlcom:
they bloats up the markdown source

qlcom:
https://github.com/rust-unofficial/awesome-rust/pull/964

palfrey:
Short answer: no
Long answer: no, because they provide useful information. However it's plausible we may want to revisit the format and consider how to redo them, or maybe move to a data format of the list that we use to say generate versions with and without build badges or something. See existing discussion at #479

qlcom:
> because they provide useful information

if someone is interested in a repository, they visit it and get more information there.

qlcom:
instead of making a simple list, you are producing giant mess. don't do that.

PR #964, opened 2020-09-18
remove build badges form Readme and don't tell to add them

qlcom:
you should edit the code on `src` dir to not to check for build badge. ( i don't know much about rust )


qlcom:
https://github.com/rust-unofficial/awesome-rust/issues/963

PR #997, opened 2020-11-29
Add linfa to Machine Learning and Are We Machine Learning Yet

Martin1887:


burdges:
ML has kinda an amusing relationship with the soundness Rust provides.  In particular, there are ML algorithms that really embrace undefined behavior.  There are presumably far more ML algorithms for which soundness helps, but UX and ML are the two areas where Rust seems like less of a slam dunk, in both cases because you may care less about bugs.  Anyways, nice links.  :)

luciusmagn:
Haha, Indeed, that is just the case. However, I do think that ML has a great future ahead in Rust nevertheless.

PR #1010, opened 2021-01-14
 Adding TerminusDB to list of Rust Databases

luke-feeney:
TerminusDB's storage is implemented in Rust - I am including in the list of databases as it may be of interest to the broader Rust community.


luciusmagn:
Cheers, that absolutely makes sense. TerminusDB is also just a cool database in and of itself, so there is absolutely merit in including it!

PR #1012, opened 2021-01-21
Added new link of the holochain repo, also kept the old one

saki-osive:
The link that points to the holochain repo has been updated to point to the new repo which is actively being maintained. The link and build status of the old repo have been kept, the new repo doesn't provide build status, so it's not been included.


luciusmagn:
Oh, I see, thanks for the quick fix!

PR #1026, opened 2021-02-02
Fix TerminusDB CI link of TerminusDB storage engine

rrooij:
It moved from Travis to GitHub Actions


luciusmagn:
Ah, shucks lmao :D thank you for updating it after such a long time. Truth to be told, we will be moving to Actions soon, since Travis is in quite a sorry state, unfortunately

PR #1027, opened 2021-02-02
Brief description of update delay-timer.

BinChengZhao:
Brief description of update delay-timer.


luciusmagn:
I know that it's very ugly, but it's uncommon for a list entry to use more than one line for an entry... I'll think about this

BinChengZhao:
> I know that it's very ugly, but it's uncommon for a list entry to use more than one line for an entry... I'll think about this

Okay, I'll try to adjust it properly on my side :) .

PR #1028, opened 2021-02-03
Add mmtc

figsoda:
mmtc is a minimal mpd terminal client that aims to be simple yet highly configurable


figsoda:
I cleaned up the branch

palfrey:

PR #1046, opened 2021-03-02
Add bunt

LukasKalbertodt:
I added the entries as specified [here](https://github.com/rust-unofficial/awesome-rust/blob/master/CONTRIBUTING.md), but in the sections I edited, there was no prior "alphabetical ordering" I could figure out. So the one I added at the end, the other I put next to another library of mine. I hope that's ok.


palfrey:

bunt passes this threshold, but reinda does not. Please update this PR to remove the reinda entry.

PR #1053, opened 2021-03-12
Update TerrabaseDB to Skytable

ohsayan:
TerrabaseDB is now known as Skytable, as a result, the reference was updated


ohsayan:
Looks like this was updated. Closed

PR #1080, opened 2021-05-05
Fix GitHub build

palfrey:
This PR does a couple of things, resulting in a working Github Actions build
* It upgrades a bunch of packages, mostly to solve the `error reading DNS system conf: Error parsing resolv.conf` error that was occurring on earlier Github actions builds
* It fixes all the current bad urls in README.md
* It adds lots of caching to the Github actions build, including some fun with `always-upload-cache` and `RANDOM` so that if we manage to get some items for the results cache, but it then borks later on, we a) still store the ones we managed to get and b) if we re-run the build, can repeatedly re-store the result file and incrementally get towards a working build. I did this with this one to get a green build, and may well be needed when this gets merged.

To clarify, this isn't me intending on coming back to maintaining and managing things (at least not yet), but I saw the issues here, and wanted to at least try and sort them out. Longer-term I might be able to spend some more time back here again, but we'll have to wait and see on that.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,10 +21,13 @@ jobs:
         toolchain: stable
         override: true
     - uses: Swatinem/rust-cache@v1    
+    - id: Get random cache id
+      run: echo "CACHE_ID=$((RANDOM))" >> $GITHUB_ENV
+      shell: bash    
     - uses: pat-s/always-upload-cache@v2.1.3
       with:
         path: results/results.yaml
-        key: results-${{ hashFiles('Cargo.lock') }}-${{ hashFiles('README.md') }}-${{ github.sha }}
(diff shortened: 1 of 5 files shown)


luciusmagn:
Thanks, I will check this out ASAP

palfrey:
> Thanks, I will check this out ASAP

How's this going?

replicadse:
From what I can tell, this change looks great. The caching is also quite interesting. Builds have been broken for quite a while and the work on the bad URLs is awesome.

palfrey:
@luciusmagn As this looks like it's a definite improvement over the current situation, unless you've got any issues with it over the next 48 hours I'm going to merge it.

luciusmagn:
Sorry, I was a bit busy, but I am no longer.

I have taken a look and it looks very good, so I'm going to expedite the process, haha

PR #1084, opened 2021-05-21
Add fancy-regex to text processing

zengxs:
[fancy-regex/fancy-regex](https://github.com/fancy-regex/fancy-regex) is a regular expressions implementation designed to support a relatively rich set of features such as look-around and backtracking.


replicadse (approved):
good catch!
looks good 👍🏽 
@luciusmagn

PR #1090, opened 2021-05-29
added frs

TheAlgorythm:
[frs](https://crates.io/crates/frs) is a project of mine to rename files with RegEx patterns.


palfrey:

TheAlgorythm:
I understand you prefer stable and mature projects. Three metrics, which might represent these goals conceivably even better are [crev reviews](https://web.crev.dev), clippy warnings and test cases. But I think only the crev reviews are viable as the other metrics would require looking at the code. (I know that frs has no review)

I understand that you need metrics, but it's a bit sad that young but mature projects don't have a chance.

palfrey:
> I understand that you need metrics, but it's a bit sad that young but mature projects don't have a chance.

Two problems there: firstly, I (or other maintainers) would need to judge "mature" projects. Personally, this would for example result in me rejecting every blockchain project as I think they're all not even slightly mature. This would be deeply unfair, and not representative of the opinions of the wider dev community.

Secondly, the maintenance overhead of maintaining this project is high enough as it is, and we badly needed to reduce it a *lot* to have any chance of keeping up with it and my not burning out on it again. I'd be happy with crev reviews, but any of the other items there are a lot of effort to review per-project, and very subjective.

PR #1091, opened 2021-05-30
Added a resource for Rust questions under section "Resources"

abiphull:


palfrey:
Why isn't this just https://stackoverflow.com/search?tab=votes&q=%5brust%5d&searchOn=3 ?

PR #1093, opened 2021-06-05
Add kalker

fluunke:
Adds kalker (https://github.com/PaddiM8/kalker) to applications.


replicadse (changes requested):
good catch
Could you also add the build badge? As far as I could see it uses GH Actions (Release workflow)

fluunke:
Done :)
Hope I did that right

replicadse (approved):
Yep, looks great! Thanks :)
@luciusmagn

PR #1096, opened 2021-06-12
Update README.md

ippan:
add https://github.com/ippan/clover - a scripting language


luciusmagn:
The error in this build is unrelated, so nothing to worry about.

PR #1097, opened 2021-06-13
Fix all the other issues with README

palfrey:
Fixes everything new since #1080 (as new things have cropped up since that got merged)


palfrey:
@luciusmagn ?

luciusmagn:
Alright, looks good. I wonder if we should maybe have a unified solution for these special regexes, but I suppose that would only have merit when more pop up

PR #1099, opened 2021-06-13
Remove Travis entirely

palfrey:
Needs #1097 for build fixes

As the Github build is now way faster and more reliable


luciusmagn:
a'ight, neat!

PR #1100, opened 2021-06-14
Added PDBTBX

nonnominandus:
_Note: I am the main author of the proposed project_

PDBTBX is a library for handling PDB/mmCIF files used for saving 3D models of proteins. These 3D models are very common in [structural bioinformatics/biology](https://en.wikipedia.org/wiki/Structural_biology).


luciusmagn:
Damn, this is really cool. Have you considered using your library to build like a rusty protein viewer?

The failure is unrelated to you, so no need to worry about that here

nonnominandus:
I was considering that but thought it would be quite hard. If you are interested or can help me thinking how to approach this problem I would certainly reconsider.

PR #1101, opened 2021-06-18
Fixes errors with pdbtbx and prost

palfrey:
Fixes the issues from #1100  as well as prost rename.

### DIFF
--- a/README.md
+++ b/README.md
@@ -784,7 +784,7 @@ See also [About Rust’s Machine Learning Community](https://medium.com/@autumn_
 
 ### Bioinformatics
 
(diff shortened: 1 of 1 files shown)


palfrey:
@luciusmagn After this gets merged we should hopefully be able to enable the "Rust build must be green to merge" and "branch must be up-to-date with master" (the latter needs the former) in Settings (see https://github.com/rust-unofficial/awesome-rust/settings/branch_protection_rules/15609230) and stop the issues from #1096 happening again as the errors were unrelated, but if master had been merged it would have had some related errors!

palfrey:
As this fixes the build @luciusmagn I'm going to merge it

Issue #479, opened 2018-06-18
Awesome-rust Renaissance (Rework of the list)


luciusmagn:
## The Premise
Awesome-rust is a very old list. So old, in fact, that it even
predates Rust v1.0. Even better, `awesome-rust` existed before Cargo
was released. Over the years, @kud1ing's tiny list grew into the
massive collections it is now. It doesn't matter what you need to do,
we probably have an entry for it.

Today, we have roughly 70 categories containing hundreds of projects.
Sadly, we also have a lot of historical baggage and this list is in
dire need of a review.

## The Problem
There are projects in the list that are unmaintained or straight up broken.
However, the fact that they are unmaintained isn't that much of a problem,
so long as they weren't left in a major WIP state.  

Another problem this list suffers from is inconsistency. Some projects have
descriptions, some don't. Some categories order by author, other by crate name,
another ones are a mess, where by we order by "what we have".

Last mildly annoying issue is that the Travis badges are slightly bigger than
lines, so as they load the content gets continuously stretched. If one opens
a link to some part of the list, they will probably end up in a whole different place
of the list.

## The Solution
I am a firm believer that it's about the time to do something about it.
Of course, the task is bigger, but these should be the main points:

- Implementation of a new, uniform format. The goal is to improve readability,
  consistency and the outlook of the list
- Careful review of all entries, removing those that are either broken, have been
  superseded by another project\* or left unmaintained in a heavy WIP state.
  It is paramount that we don't make a mistake here, so emphasis on the careful
  part.
- Filling in info for valid entries that lack either description, link to https://crates.io
  (if applicable), link to the repository (ditto) or a badge
- Review of the validity of categories (we had some issues with ambiguity
   quite some overlapping in the area of UI), turning some into subcategories
   of others, if needed
- On the topic of categories, we should also have a uniform ordering guideline,
  but whether we order by author or project name is up for discussion at the
  moment
- Doing something about the Travis badges. I am hoping that something like
  https://shields.io might help solve the problem with their custom badges,
  but more research needs to be done in the area. One solution is to remove
  them completely, but in doing so we lose some valuable easily accessible
  information. The best solution would be badges that wouldn't stretch the lines.
- Update the contribution guide to include all the relevant information, maybe
  add some policies to it. `awesome-rust` should definitely follow (and quietly does,
  I believe) the Rust Code of Conduct, so I am talking more about policies
  on what content we accept, what not. 

## The Plan
[...]

PR #1118, opened 2021-07-11
Add Svetlitski/fcp under 'System tools'

Svetlitski:
I'd like to humbly suggest my own tool, [`fcp`](https://github.com/Svetlitski/fcp), for inclusion on this list. It's a modern replacement for `cp` which is heavily focused on performance and therefore *much* faster (i.e. anywhere from 1 to nearly 3 orders of magnitude faster than `cp` for the same workload).


Svetlitski:
Having read through #479 I now realize that more items is probably the last thing this list needs, and thus am closing my PR.

PR #1131, opened 2021-08-05
Add some configurations and fix broken links

williamdes:


williamdes:
After merge I will open a PR to remove trailing whitespaces

.gitattributes
```diff
@@ -0,0 +1,5 @@
+/.github export-ignore
+/.gitignore export-ignore
+/CONTRIBUTING.md export-ignore
+/.markdownlint.json export-ignore
+/.editorconfig export-ignore
```
palfrey, line 5: Why the need to disable these for archives?

.markdownlint.json
```diff
@@ -0,0 +1,10 @@
+{
+    "default": true,
+    "MD004": false,
+    "MD007": false,
+    "MD013": {
+        "line_length": 1000
+    },
+    "MD024": false,
+    "MD032": false
+}
```
palfrey, line 10: I'm not totally sure about this list (MD032 for example looks maybe should be enabled, but haven't tested yet). Also, if this config is here but not being run in CI, even if we fix things right now with it, it'll get degraded over time. Can you add something to the CI to run the lint?

.gitattributes
williamdes: It is not "needed" but a nice to have
People do not care about dev files when they download your repo `git archive`: https://github.com/rust-unofficial/awesome-rust/archive/refs/heads/master.zip

.markdownlint.json
williamdes: Sure I can add a CI check for it, I disabled all that was the actual style in the file
My proposal is to add a CI check and after merge fix whitespaces and the Markdown lint issues, sounds good ?

williamdes:
Here it is 🚀 
A failure: https://github.com/williamdes/awesome-rust/runs/3273769100?check_suite_focus=true
And now success

Issue #981, opened 2020-10-09
Project issues


palfrey:
This is to some extent a continuation of #479, but also some other items. I'm interested in particular in @luciusmagn and @kud1ing's thoughts, but others are also welcomed

So, I'm feeling a bit burnout on this project, and as such will be switching off notifications on new PRs and won't be reviewing them for at least a bit. A lot of my issues come down to the following problems:

* Quality of repos: we don't really have any. It's called "Awesome Rust", and TBH our quality control (for which I'm as guilty as any) is not great. This results in a really long list, which is nice in certain ways, but I'm wondering if quality rather than quantity would be a better idea. I'm leaning towards more mechanical options of quality (e.g. Github stars or crates.io downloads), mainly to try and a) provide a consistent process and b) avoid the maintainers here being the judges of such matter, and using existing community metrics instead. If we do that, the thresholds should be clearly and publicly documented. There will always be popular projects that don't fit into any of the metrics, and in those cases we first need to find a new metric.
* Quality of list items: Our formatting is a bit of a mess. There's kinda a style, but it's not enforced, and there's kinda alphabetical ordering, but also not enforced. I think we need to have a fixed style, and enforce it. Main goal here is to reduce maintainer time i.e. make it easy to do the right route, so this should be an automatic thing. Probably a CI check. Maybe fits in with the JSON data concepts that were being floated in #479.
* Time spent on PRs: We need something like the Github pull request templates plus saved replies. The template should have a bunch of common issues to tick off, and there should be an expectation that if you open a PR and you've ignored one or more of those, it's getting closed pretty damn quick. It's somewhat of a jerk move, but from a "maintainer time" perspective, it's a win, and I think needed.

Thoughts anyone?

kud1ing:
Hi,

> So, I'm feeling a bit burnout on this project

I am sorry to hear that. But I can relate to this very much.

Regarding:
* Quality of projects to link to: i am all for quality over quantity. This is a quite subjective and difficult topic. Especially when it comes to inactive but important projects.
* Quality control: i've always felt that a cleanup script is necessary to enforce a common style (to fix ordering, whitespace, dashes etc.).

Thanks for your work so far.

To be honest, i don't plan to invest much time in this project anymore. Maybe it's time to look for maintainers again or let it be?

quintrino:
[...]

PR #738, opened 2019-10-20
Games: Remove discontinued zoc

trivikr:
Details: https://github.com/ozkriff/zoc

PR #1141, opened 2021-08-10
Popularity filtering

palfrey:
In line with some of the stuff I'd talked about in #981, and with the explicit objective of making it easier to maintain this list going forwards, I've implemented here a popularity filter for the list. The first time it sees an entry, it tries to get github stars and cargo downloads for that entry. If they exceed some documented thresholds (50 and 2000 respectively) it stores that so we never have to check that project for those numbers again (as they almost never go down, or at least not much).

I've removed all the projects that didn't achieve this threshold (and added a few cargo links to help some that should have been allowed). There's also a `POPULARITY_OVERRIDES` list at the top of the main.rs that can be used to add those projects that are that popular but the current detection mechanisms don't pick up.

If merged, this would result in us being able to reject probably about half of the current PRs as they are below those thresholds, and probably once people fix the other ones, allow us to fairly rapidly decide if a project should be added, which should massively speed up future PR review.

@kud1ing @luciusmagn Thoughts?


README.md
```diff
@@ -1636,7 +1436,6 @@ See also [Are we web yet?](https://www.arewewebyet.org) and [Rust web framework
 
 A registry allows you to publish your Rust libraries as crate packages, to share them with others publicly and privately.
 
-* [Cloudsmith :heavy_dollar_sign:](https://cloudsmith.com/cargo-registry/) — A fully managed package management SaaS, with first-class support for public and private Cargo/Rust registries (plus many others). Has a generous free-tier and is also completely free for open-source.
```
palfrey, line 1639: I'm not totally sure about this one. Evaluation of commercial things in this way is hard, but https://github.com/cloudsmith-io has only a handful of stars and no Rust code, so I'm kinda leaning towards the delete option.

kud1ing:
I think automatisation is important to reduce the maintainer's load.
I also think it's very important that the content stays high quality. This means saying no to additions and removing entries. This will always be subjective and controversial.

Your attempt seems like a good start. After only a cursory browsing, some random thoughts:
* maybe be more generous with some underrepresented and/or educational niches like games
* maybe park the removed entries somewhere so that the information is not lost.
* removed entries are likely to be added again by someone in the future. Maybe check the PRs with the parking lot of previously removed entries.

Thank you for keeping this alive.

palfrey:
> I also think it's very important that the content stays high quality. This means saying no to additions and removing entries. This will always be subjective and controversial. See e.g. [#738 (comment)](https://github.com/rust-unofficial/awesome-rust/pull/738#issuecomment-573468900)

In a way, the removals are easier because we've already got a well-defined line on them (and low-quality removals are less of a problem) and even if that line is controversial, it's fairly clear, which is good from a "maintainer time" perspective.

>     * maybe be more generous with some underrepresented and/or educational niches like games

Agreed. There was definitely some low-quality ones there, but also two projects with 40+ stars who got removed. Obviously there's always going to be projects who just miss whatever bar we put in, but in this particular case it's worth adding in some lowered thresholds for some areas. I'd say that overall this shouldn't be done for most things, and if an area gets up the numbers then the bar should be reset back, but I'll do this for games and see if there's any other notable areas with multiple "almost good enough" items.

>     * maybe park the removed entries somewhere so that the information is not lost.
>     * removed entries are likely to be added again by someone in the future. Maybe check the PRs with the parking lot of previously removed entries.

I'm not sure about this one. So, firstly if they get re-added and still fail, then they should get rapidly closed, so I'm less concerned about that. I'm also wondering where to sensibly park them? The option I can think of is some sort of extra file in the repo, and that almost seems mean-spirited ("here's a list of projects that aren't good enough" kinda thing). If you've got some other idea that aids discovery I'd love to hear it.
 
>     * maybe automatically remove the `foo/bar` namespace so that is only `bar` for very popular entries.

Worth looking into, but more of a "fix/unify display of items" issue (which I might get to eventually)

> Thank you for keeping this alive.

I like this list, and I'm hoping that adding in things like this will reduce the maintenance burden and so reduce the odds of me burning out again. Maybe we can even get some new maintainers eventually :)

README.md
lskillen: (I work there) Cloudsmith are supporting the ecosystem, though. We were the first service to support Cargo/Crates (outside of Crates.io) itself, as you can see here: https://cloudsmith.com/blog/worlds-first-private-cargo-registry-w-cloudsmith-rust/ - plus we've contributed to the ecosystem itself, and have customers (such as Embark, several -ifys, etc.) that also contribute. Although we're not personally built on Rust, it's something that might happen (e.g. static single dependency CLI).

palfrey: That's fair. I think this falls under the "Rust doesn't have much of this thing, so we've got slightly different standards for that area" criteria that's been noted for some other bits. Also, I can now document this in the comments about why we're overriding so that your entry gets kept in :)

I've re-added you.

palfrey:
In the absence of other feedback, I'm going to merge this on Monday 16th (i.e two days time) and start closing out PRs after that that fail the criteria.

README.md
lskillen: Very much appreciated, Tom, thank you! We'll continue to support the Rust (and other) ecosystems as best we can. :-)

PR #1184, opened 2021-11-01
Added stalwart-labs/mail-parser to Email category.

mdecimus:
Hi,

We just released a fully-featured e-mail parsing library for Rust and would love to have it listed on awesome-rust.

Thanks!


palfrey:

mdecimus:
Thanks for the feedback. The library was published yesterday and it has just reached 57 stars, would that be enough to be added or do you also require 2000 downloads in addition to the 50 stars? Thanks.

palfrey:
> Thanks for the feedback. The library was published yesterday and it has just reached 57 stars, would that be enough to be added or do you also require 2000 downloads in addition to the 50 stars? Thanks.

We just need any one to exceed the threshold. That's a lot higher than it was yesterday!

PR #1185, opened 2021-11-02
Add Free PO file editor

goranluledzija:
Hey there,
I have added a Free PO file editor, it is an online tool for translating PO files for Rust localization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,7 @@ See also [A comparison of operating systems written in Rust](https://github.com/
 ### Text editors
 
 * [amp](https://amp.rs) — Inspired by Vi/Vim. [![build badge](https://api.travis-ci.org/jmacdonald/amp.svg?branch=master)](https://travis-ci.org/jmacdonald/amp)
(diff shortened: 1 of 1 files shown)


palfrey:
I'm not seeing any Rust link here (other than there's probably several rust crates that deal with PO files). Is the tool written in Rust, or any other links to it?

goranluledzija:
PO files are used by Gettext tool that extracts translation texts from the source code.
Rust is one of programming languages that rely on Gettext for localization, take this for example https://lib.rs/crates/gettext-rs

The PO file editor is useful for anybody that relies on Gettext for localization.

palfrey:
I agree that there are Rust crates that use PO files. What I'm not seeing is either a) a project written in Rust or b) a project of general interest to the overall Rust community, as opposed to the subset of people using Gettext. If we add projects where "someone who uses crate X might like this tool", we could add anything, so we need to be pickier than that. I'm closing this request, but feel free to reopen if you've got further evidence of this being of use to the wider Rust community.

goranluledzija:
Ok, but I think it will benefit majority of Rust developers since Gettext is preferred localization approach in Rust.

PR #1191, opened 2021-11-16
add revolt

azzamsa:


palfrey:
This doesn't appear to be a Rust project, and so unsuitable for this list. A link to [Delta](https://github.com/revoltchat/delta) specifically would be suitable though.

azzamsa:
> This doesn't appear to be a Rust project, and so unsuitable for this list. A link to Delta specifically would be suitable though.

done.

PR #1193, opened 2021-11-21
Add Hands-on Rust

thebracket:
Hands-on Rust (https://pragprog.com/titles/hwrust/hands-on-rust/) is my book about learning Rust, through the medium of creating games. It should meet the popularity requirements: it's been in the top 5 best-sellers at pragprog.com (publisher) for 50 of the last 53 weeks, and is usually in the top 20 best selling books in the "C Language" category on Amazon.com (this changes every 5 minutes, so it's a tough one to measure - see https://www.amazon.com/dp/1680508164#SalesRank ).


palfrey:
Agreed, that seems like a decent level of popularity. Thanks for providing the data!

PR #1204, opened 2021-12-17
Adding  libmdbx-rs fork

parasew:
Rust bindings for MDBX —  fork of mozilla/lmdb-rs with patches to make it work with libmdbx.. Fast, compact, powerful, embedded, transactional key-value database, with permissive license.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,7 @@ See also [Rust — Production](https://www.rust-lang.org/production) organizatio
 * [TerminusDB](https://github.com/terminusdb/terminusdb-store) - open source graph database and document store [![Build Status](https://github.com/terminusdb/terminusdb-store/workflows/Build/badge.svg?branch=master)](https://github.com/terminusdb/terminusdb-store/actions)
 * [tikv](https://github.com/tikv/tikv) — A distributed KV database in Rust [![Build Status](https://ci.pingcap.net/job/tikv_ghpr_test/badge/icon)](https://ci.pingcap.net/job/tikv_ghpr_test/)
 * [WooriDB](https://github.com/naomijub/wooridb) - General purpose time serial database inspired by Crux and Datomic.
+* [libmdbx-rs](https://github.com/vorot93/libmdbx-rs) Rust bindings for MDBX —  fork of [mozilla/lmdb-rs](https://github.com/mozilla/lmdb-rs) with patches to make it work with [libmdbx](https://github.com/erthink/libmdbx). Fast, compact, powerful, embedded, transactional key-value database, with permissive license. 
 
 ### Emulators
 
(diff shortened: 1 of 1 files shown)


palfrey:

parasew:
[Libmbdx](https://github.com/erthink/libmdbx) has 931 stars on github. I also would define [lmdb-rs](https://github.com/mozilla/lmdb-rs) from Mozilla as relevant (although it has only 37 stars). I agree that the threshold is a problem, especially because downloads without releases on Github are not counted. The overall metrics of the mdbx-sys crate (all versions combined) shows[ over 1,933 downloads accumulated](https://crates.io/crates/mdbx-sys) and the [reverse_dependencies of mdbx-sys](https://crates.io/crates/mdbx-sys/reverse_dependencies) (dependents) lists [heed](https://crates.io/crates/heed) which itself has more than 22k downloads. Does that qualify?

LMDB is quite relevant in my opinion (and also objectively), as it has been [originally developed for the  OpenLDAP project](https://www.symas.com/lmdb) - and already has an entry in this list. LMBDX is a faster ("one of the fastest compact embeddable key-value ACID database without WAL"), with being [almost 30% faster than LMDB in CRUD benchmarks.](https://erthink.github.io/libmdbx/intro.html) and has a few other improvements (see the description).

palfrey:
> [Libmbdx](https://github.com/erthink/libmdbx) has 931 stars on github. I also would define [lmdb-rs](https://github.com/mozilla/lmdb-rs) from Mozilla as relevant (although it has only 37 stars). I agree that the threshold of mdbx-sys (libmdbx-rs) is a problem, especially because downloads without releases on Github are not counted. The overall metrics of the mdbx-sys crate (all versions combined) shows[ over 1,933 downloads accumulated](https://crates.io/crates/mdbx-sys) and the [reverse_dependencies of mdbx-sys](https://crates.io/crates/mdbx-sys/reverse_dependencies) (dependents) lists [heed](https://crates.io/crates/heed) which itself has more than 22k downloads. Does that qualify?

The mozilla numbers are irrelevant, as you're a fork of them. If they'd declared you the One True Fork, even then not really, because stars between those sorts of things tend to be a lot of the same people. https://crates.io/crates/mdbx-sys has below the threshold, so doesn't count (2000 is a pretty low threshold for a popular library TBH, so I'm only counting things explicitly over that). The heed one has mbdx as a optional dep, and by default disabled in favour of https://crates.io/crates/lmdb-rkv-sys with 289k downloads, so not really getting there either. Also, reverse deps tend to in general result in an increase in the crate downloads, so it's not worth paying attention to them as if they are actually resulting in more uses of your crate, your downloads should go up!

The crate numbers are close though, so come back with this one when it gets there. I suspect only a matter of a few weeks, but hard to tell from the graph.

> LMDB is quite relevant in my opinion (and also objectively), as it has been [originally developed for the OpenLDAP project](https://www.symas.com/lmdb) - and already has an entry in this list. LMBDX is a faster ("one of the fastest compact embeddable key-value ACID database without WAL"), with being [almost 30% faster than LMDB in CRUD benchmarks.](https://erthink.github.io/libmdbx/intro.html) and has a few other improvements (see the description).

I am deliberately paying no attention to any of this, as the point of the metric-based decisions is to avoid subjective maintainer-specific choices on how relevant a project is. Also, none of that is Rust-specific, and just because it's a faster thing in other language implementations doesn't mean it's of interest to here necessarily.

parasew:
Hi and thanks for the answer. I just saw that mdbx-sys seems to have hit the 2000 download mark (accumulated across versions), so I guess it qualifies now? Should I re-submit? Thanks, Matthias

palfrey:
No, I can just re-open this. Please fix the issues flagged in the build steps.

PR #1213, opened 2022-01-03
Add's r-zenine/sam

r-zenine:
Hi, 

I would like to propose to add a tool I am writing in the list. 
It's quite stable and used by myself and many of my colleagues on our day to day work. 

Disclaimer: I am the author of the tool, therefore I am biased towards thinking it is a useful tool :). I hope you will find it useful too.


palfrey:

r-zenine:
Hello @palfrey , 

That's fair enough. I was not aware of those guideline. 
I'll resubmit if my project reaches those thresholds. 

Thank you

palfrey:
It's in the [CONTRIBUTING.MD](https://github.com/rust-unofficial/awesome-rust/blob/master/CONTRIBUTING.md)

>- is the entry valuable to people trying to get things done in Rust?
>    * In order to make this objective, the entry needs to either have at least 50 stars on Github, 2000 downloads on crates.io, or an equivalent level of other popularity metrics (which should be specified in the PR). The maintainers of this repo are not responsible for making your project popular, only for making more people aware of those projects. We don't want to have to pick and choose favourites, and so are using metrics like this to make our lives easier as maintainers.

PR #1214, opened 2022-01-03
Fix description and links for `cucumber` crate

tyranron:


palfrey:
There was both a 301 error with the docs link - I'd suggest linking to crates there, not docs, especially given you're using the crates badge; and now the testing section got reorganised by https://github.com/rust-unofficial/awesome-rust/pull/1215 so this needs to resolve the conflicts with that.

PR #1215, opened 2022-01-03
Subdivide testing section

mig4:
There are a couple of different categories under the testing section.
This subdivides that section to make things easier to find.

Also add link to #test crates list. This seems almost as popular as
#testing and not all projects have both tags.


palfrey:
Nice, thanks!

PR #1223, opened 2022-01-21
Improve vscode plugin section

guidsdo:
I think this can help people a bit more when learning Rust and using vscode :-)


guidsdo:
I've removed the extension that didn't have enough stars :)

palfrey:
> I've removed the extension that didn't have enough stars :)

Not enough stars, because it's not a Github repo :) OTOH, it does have 860K downloads, so that's more than enough, but not something we automatically count. If the URL gets added to `POPULARITY_OVERRIDES` though (see https://github.com/rust-unofficial/awesome-rust/blob/master/src/main.rs#L53) it'll work fine

guidsdo:
> > I've removed the extension that didn't have enough stars :)
> 
> Not enough stars, because it's not a Github repo :) OTOH, it does have 860K downloads, so that's more than enough, but not something we automatically count. If the URL gets added to `POPULARITY_OVERRIDES` though (see https://github.com/rust-unofficial/awesome-rust/blob/master/src/main.rs#L53) it'll work fine

Haha thanks! Unfortunately not many people vote for extensions anyway so counting the vscode stars isn't useful to add anyway indeed :)

PR #1225, opened 2022-01-22
Revert removal of cpc

probablykasper:
Tried to figure out why it was removed, but in the history it keeps getting added and removed over and over, so I'm guessing it was a mistake?


palfrey:
It got removed by https://github.com/rust-unofficial/awesome-rust/pull/1141 and given the threshold for popularity was set to 50 stars and it's now got 53, I'm guessing it was a bit below at the time. Happy to have it back though!

PR #1227, opened 2022-01-23
add rsocx

b23r0:


palfrey:
There's a travis config, but https://app.travis-ci.com/github/b23r0/rsocx appears to be lacking in recent builds. Is that the right build badge?

b23r0:
> There's a travis config, but https://app.travis-ci.com/github/b23r0/rsocx appears to be lacking in recent builds. Is that the right build badge?

replace to github action badge

PR #1228, opened 2022-01-26
Added 6 videos to Presentations

stjepanjurekovic:
Hi,

Stjepan from Manning here. i thought these six videos (including Manning's Rust conference) might be a good match for your list. Thank you for considering it.

Best,


palfrey:
That's quite a few videos, plus at least one that isn't properly linked. Can you make a playlist for Manning Rust content and we can just link that please?

Also this has conflicts with the current master README, so those will need resolving.

palfrey:
Closing temporarily while we await a fix for the conflicts and lack of playlist.

PR #1233, opened 2022-02-09
Add totp-rs

constantoine:
As of time of writing, totp-rs has been downloaded ~4.6k times. I think it would be a good fit here, I just hope I didn't put it in the wrong category

### DIFF
--- a/README.md
+++ b/README.md
@@ -731,11 +731,11 @@ See also [About Rust’s Machine Learning Community](https://medium.com/@autumn_
 
 ### Authentication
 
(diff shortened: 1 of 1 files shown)


palfrey:
FYI, the sort suggested putting this at the top of the list

palfrey:
> @palfrey Ah, my bad, I sorted i by crate name, not username
> 
> Should I remove my username to keep it sorted, or put it on top?

On top is fine. We're pretty ad-hoc on the naming, and that name is fine.

PR #1244, opened 2022-03-12
Validate that links are up once a week

simeg:
Every Sunday at midnight this Github action will validate that all
http/https links responds with 200. If a link has been permanently
removed (or just moved) it can be removed/changed which leads to a
higher quality list.

This check is already in place in the mac-setup[0] repo and is working
well. It won't prevent any PRs from getting merged, and optionally a
badge can be added to the README to highlight the link status (see
mac-setup repo).

[0]: https://github.com/sb2nov/mac-setup/

---

Hi! I'm curious what you think of my suggestion. I think this will increase the quality of this nice repo! 🙂

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -1,0 +1,36 @@
+name: Links
+
+on:
+  schedule:
+    # Run at 00:00 UTC every Sunday
+    - cron: '0 0 * * */7'
+
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Find markdown files
+        shell: bash
+        run: |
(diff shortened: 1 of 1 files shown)


palfrey:
https://github.com/rust-unofficial/awesome-rust/blob/master/.github/workflows/rust.yml already checks this

simeg:
@palfrey What part of the workflow does this exactly? I can't find it 🙂 This is not linting, this is checking if the URLs are up

palfrey:
Yup. Also, very importantly for this repo we cache the prior results. Even with a Github token set in the headers, we still run into rate limits with "429 Too Many Requests" for Github links. To handle this, the link checker checks daily, starts from "oldest check first" and doesn't worry about 429's (although doesn't mark them as updated). Everything not on Github gets checked daily, and everything on Github gets checked about effectively every 3 days. Also, for some things (mostly stuff like 500s) we don't fail the build until it's fallen over for a week as a fair number of sites break periodically but will fix fast enough for most users and come back.

PR #1246, opened 2022-03-18
Add terminal typeracer to the list of games

DarrienG:


DarrienG:
Aw man, it looks like the CI job only checks for GitHub stars :( I have 50 GitLab stars https://gitlab.com/ttyperacer/terminal-typeracer which is probably every GitLab user. Perhaps that's ok?

palfrey:
> Aw man, it looks like the CI job only checks for GitHub stars :( I have 50 GitLab stars https://gitlab.com/ttyperacer/terminal-typeracer which is probably every GitLab user. Perhaps that's ok?

Yes, that works. I haven't implemented checks for Gitlab, but if you can add your repo URL to https://github.com/rust-unofficial/awesome-rust/blob/master/src/main.rs#L43 with a comment noting about the stars. Basically, most of our repos are Github, so that's worth doing automatically, but Gitlab just hasn't had enough yet for it to be worth it (or a Gitlab project has a high downloads Cargo package most of the time, so it passes on that instead).

PR #1249, opened 2022-03-23
Adding a new Simulation section under Applications

Alfred-Mountfield:


README.md
```diff
@@ -299,6 +300,10 @@ See also [A comparison of operating systems written in Rust](https://github.com/
 * [ripasso](https://github.com/cortex/ripasso/) — A password manager, filesystem compatible with pass
 * [rustscan/rustscan](https://github.com/RustScan/RustScan) — Make Nmap faster with this port scanning tool [![build badge](https://github.com/RustScan/RustScan/workflows/Continuous%20integration/badge.svg?branch=master)](https://github.com/RustScan/RustScan/actions?query=workflow%3A%22Continuous+integration%22)
 
+### Simulation
+
+* [hEngine](https://github.com/hashintel/hash) - A Rust-implemented computational simulation engine, supporting large-scale agent-based modelling, with simulation logic written in JavaScript and Python.
```
palfrey, line 305: The repo appears to be for something called HASH, not hEngine?

Alfred-Mountfield: Hi there, the link is to the HASH (company) monorepo which contains the hEngine. The hEngine is directly signposted from the root README so I thought it'd be better to point at the repository rather than a subfolder in case the folder structure changes and invalidates the link.

Is that problematic, should I change it to something?

palfrey: Can you please link to the subfolder instead, as that's less confusing for day-to-day users (and myself TBH!) and I'm assuming won't be moving _that_ often...

Alfred-Mountfield: Yep no worries, will just try to get a process in place to remember to update it if the folder structure does change

Alfred-Mountfield: Alternatively, would it be less confusing if I titled it as HASH Engine here? Which is the full name anyway. Otherwise yeah I'm happy to link the subfolder and try and get a process in place to remember to update it if the folder structure does change

palfrey: Only hEngine appears to be Rust, so please link to that. The rest of the project isn't relevant to this repository.

Alfred-Mountfield: Yep that's fair :) have updated!

PR #1251, opened 2022-03-29
Add expectrl to cli/pipelines

zhiburt:
Hi there

I am not sure about the order because seems like this section doesn't preserve it?


palfrey:
> I am not sure about the order because seems like this section doesn't preserve it?

Order was correct, as there's part of the CI that checks. It's just by the full name of the project including your username.

PR #1256, opened 2022-04-07
Add Theseus OS to Operating Systems

kevinaboos:
Added Theseus OS to the Operating Systems section.

Thanks for maintaining this awesome list!


kevinaboos:
Hmm, it looks like the CI failure is due to the tool assuming that a redirect `301` is an actual error. Both links work, though.

kevinaboos:
@palfrey the link check linter ought to pass now, thanks.

palfrey:
> @palfrey the link check linter ought to pass now, thanks.

Thanks. We've had too many things that have a 301 only to later disappear, so we want to to make sure we've got the updated URL for things!

kevinaboos:
Totally understand. Thanks!

PR #1266, opened 2022-04-18
Add AWS SDK + Qovery in Cloud section

deimosfr:


palfrey:
Qovery isn't AWS specific so should be at the general cloud level not the AWS section. Also the alphabetical ordering needs fixing (see build failure)

deimosfr:
> Qovery isn't AWS specific so should be at the general cloud level not the AWS section. Also the alphabetical ordering needs fixing (see build failure)

My bad, fixed

Issue #419, opened 2018-02-21
Change format for list entries


memoryruins:
As awesome-rust grows, there are a few tweaks that may improve readability, consistency, and load times. Before making changes en masse, it would be great to hear anyone's thoughts and a green light from maintainers.

* e.g current format (for half the current list. currently, not all entries include links to crates)
  * [kbknapp/clap-rs](https://github.com/kbknapp/clap-rs) [[clap](https://crates.io/crates/clap)] — a simple to use, full featured command-line argument parser [<img src="https://api.travis-ci.org/kbknapp/clap-rs.svg?branch=master">](https://travis-ci.org/kbknapp/clap-rs)
  * [docopt/docopt.rs](https://github.com/docopt/docopt.rs) [[docopt](https://crates.io/crates/docopt)] — a Rust implementation of [DocOpt](http://docopt.org) [<img src="https://api.travis-ci.org/docopt/docopt.rs.svg?branch=master">](https://travis-ci.org/docopt/docopt.rs)  
  * [TeXitoi/structopt](https://github.com/TeXitoi/structopt) [[structopt](https://crates.io/crates/structopt)] — parse command line argument by defining a struct [<img src="https://api.travis-ci.org/TeXitoi/structopt.svg?branch=master">](https://travis-ci.org/TeXitoi/structopt)
  * [killercup/quicli](https://github.com/killercup/quicli) [[quicli](https://crates.io/crates/quicli)] — quickly build cool CLI apps in Rust [<img src="https://api.travis-ci.org/killercup/quicli.svg">](https://travis-ci.org/killercup/quicli)
* proposed changes
  * present the name of the project/library first, which brings it closer to the name used in a cargo.toml in many cases. keep it linked to the repository.
  * remove build passing icon. the page makes many requests to travis for badge SVGs. especially from a slower connection, the  loading of the badges move the page as they affect the vertical spacing of the lists. if a user jumps or scrolls to the bottom of the page upon opening, they may need to wait before the page stays still. this change loses a nice splash of color the list, but reduces load times and prevents the unaligned badges slowing and littering the page.
  * move crates and build status links to a sub-list. name links as crates.io,  build status, etc. this gives the description breathing room and allows us to include additional information/links on the sub-list, e.g. a direct link to the docs.
* proposed format
  * [**clap**](https://github.com/kbknapp/clap-rs) — a simple to use, full featured command-line argument parser.
[...]

PR #1275, opened 2022-04-28
Add star badge of per github repo

ColinZeb:
Add a star badge to help someone check quicly


palfrey:
So I'm really not sure about this one. There's been prior complaints about the load times for the page already, and this automatically adds a chunk to that. Also, it makes adding a new entry more complicated, and those are already fairly inconsistent already. Related to that, the stars should probably be at the end of the entry along with all the other badges.

If you want to talk about this in the context of the wider list reformat issues (see https://github.com/rust-unofficial/awesome-rust/issues/419 and https://github.com/rust-unofficial/awesome-rust/issues/479 in particular), that's great, but it'll need a lot more work than just an initial "add all the badges" PR.

PR #1279, opened 2022-05-10
Added Chemistry and Physics

JASory:
Molecular, atomic and subatomic modeling libraries for chemistry, nuclear and plasma physics. Due to the relatively low popularity of crates in scientific research, none of them strictly meet the requirements for notability, however they are the best (and generally most downloaded/starred) in their field.


palfrey:
I'm willing to drop the stars required to 30 for both, but not to the ~10 that would be required for two of the physics crates. You can add the overrides to https://github.com/rust-unofficial/awesome-rust/blob/master/src/main.rs#L26 and you'll also need to fix the linting issues (see build logs).

PR #1280, opened 2022-05-15
Add orm prisma client rust

skimhugo:
Prisma client rust crate let rust can use the power of prisma.


palfrey:
Should be in the ORM block not OGM as AFAIK it doesn't do graph databases?

skimhugo:
@palfrey I fix it, this file is so long which made me dizzy

PR #1299, opened 2022-06-16
Add fred to Redis clients

aembke:
Fred is a high level Redis client built for Tokio. The repository linked here is the new location, and the old repository can be found [here](https://github.com/azuqua/fred.rs).


aembke:
Looks like the build step failed due to a redirect from CircleCI. I changed the badge link to the 301 destination to hopefully fix that.

PR #1303, opened 2022-06-20
Add Wasmtime, in Libraries / Virtualization.

sunfishcode:
I'm open to ideas for putting this in a different category.

WebAssembly itself doesn't neatly fit within the existing categories.
I picked virtualization here because that's one of the things it's known
for. Similarly, Wasmtime can be used as a standalone application, but
it's more known for being used as a library.


palfrey:
It's not a bad place to go. Wasmer (which seems similar) is in the generic "Applications" but that's a bit of a dumping ground. I'd have leant towards "Applications / Virtualization" based on the README, but happy to have it in libraries instead. Thoughts @sunfishcode?

sunfishcode:
Wasmtime has a CLI "application", but it's mainly used for development. The major production use cases for it all use the library crates. So I think Library makes sense.

PR #1322, opened 2022-07-26
Add CodeCrafters & Build your own X

sarupbanskota:
* [Build your own X (CodeCrafters.io)](https://codecrafters.io) — Rust mastery, by recreating Redis, Docker, Git, and SQLite
  * [Build your own X (repository)](https://github.com/codecrafters-io/build-your-own-x) — A list of _Build your own X_ style exercises in Rust and other languages


palfrey:
I removed CodeCrafters in https://github.com/rust-unofficial/awesome-rust/commit/a42aa6b9d199cf12d1a7078725369b309e9e4719 because you got rid of the Rust-specific link. Both of those links are very generic and not really Rust specific, so they don't really belong here. If you wanted to add links for some of the Rust-specific items from the latter that would be interesting, but it's mostly non-Rust.

PR #1326, opened 2022-07-30
Add cargo-mutants

sourcefrog:
Split out a new "mutation testing" category because there are now two crates in this category, and it's a different thing from fuzzing. (See
https://github.com/sourcefrog/cargo-mutants/wiki/Compared).

Fix a typo


sourcefrog:
As you can see it has [98 stars](https://star-history.com/#sourcefrog/cargo-mutants&Date) and [1699 downloads](https://crates.io/crates/cargo-mutants).

PR #1332, opened 2022-08-10
Update socketcan-rs URL

jason-h-35:
This repo moved and now it's triggering an ❌ in the GitHub Actions for this repo.


palfrey:
Thanks for doing this. The new Travis badge isn't working though, as they haven't enabled the new repo. [I've filed a bug on this](https://github.com/socketcan-rs/socketcan-rs/issues/25) but for the moment we should probably just remove the build badge as they don't have any currently functioning CI.

PR #1334, opened 2022-08-10
Add crates.nvim to IDE -> Vim section

andriisoldatenko:


palfrey:
The sorting for this needs fixing. See the Rust build log.

PR #1343, opened 2022-08-24
Remove deprecated language server

SUPERCILEX:
https://blog.rust-lang.org/2022/07/01/RLS-deprecation.html


SUPERCILEX:
The error doesn't seem related to this change.

palfrey:
> The error doesn't seem related to this change.

Correct. Fixed in 88d27814e7af75bcd0daf39b0046aab9c4fc3f1c

PR #1409, opened 2023-01-23
Add Obliteration on Emulator list.

p0008874:
Hey babe, The new PlayStation 4 emulator dropped.


palfrey:
Thanks for the PoloDB fix :)

PR #1423, opened 2023-02-05
Add cargo-crev #HSFDPMUW

demondave:
https://github.com/crev-dev/cargo-crev


palfrey:
Out of curiosity because I've seen it on a few recent pull requests, but doesn't turn up in Google: What's HSFDPMUW?

demondave:
> Out of curiosity because I've seen it on a few recent pull requests, but doesn't turn up in Google: What's HSFDPMUW?

We have a class university where we have to contribute to open source projects. To identify our contributions we have to include the HSFDPMUW hashtag.

PR #1463, opened 2023-03-26
Add nushell/nushell

SUPERCILEX:


Israel-Laguan (approved):
Maybe try to add a build bag, which you can take from their readme
image


I confirmed it has more than 50 stars
image
And more than 2k downloads
image

PR #1466, opened 2023-03-26
Added dialoguer library for terminal prompts

StaticESC:
Was trying to find a good terminal library in which one could select an options, came across dialoguer and thought that it deserves to be mentioned here. It does something similar to this: https://imgur.com/a/90Bn9Bt It has close to 1k stars.


palfrey:
Sort order needs fixing. See build log for details

Israel-Laguan (approved):

PR #1469, opened 2023-03-31
Rust Weekly

xojoc:


palfrey:
So, this newsletter is completely new to me. 
If it was a github project we'd just count the stars/downloads, but I'm not sure what the equivalent is here. Can you provide any publicly available metrics to indicate this is a popular newsletter?

xojoc:
It's not that popular I'm afraid. I'll create a new pull request once I get more subscribers. Thanks

PR #1473, opened 2023-04-06
Add aero

rilysh:


Israel-Laguan (approved):
More than 50 stars
image

No presence on crates.io 😢 

Please move the entry up as the build asks for alphabetical order

palfrey:
The travis links were all very broken now so I've just removed them (https://github.com/rust-unofficial/awesome-rust/pull/1475), which has caused a minor conflict here. If you can fix that, we can merge.

rilysh:
Conflicts now should be fixed

PR #1474, opened 2023-04-06
Adding Zenoh Protocol to Network Programming

kydos:
Zenoh is a pure rust Pub/Sub/Query protocol used in  IoT, Edge, Robotics, Automotive, etc.


palfrey:
There's a sort issue that needs fixing (see build logs) and also a conflict because I got rid of all the broken Travis links yesterday, but that should be an easy fix.

palfrey:
Closing while we wait for the fixes.

PR #1477, opened 2023-04-07
Add cargo-udeps

ericcornelissen:
Adding [cargo-udeps](https://github.com/est31/cargo-udeps), a tool to find unused dependencies.

I put it in the ["Build system" section](https://github.com/rust-unofficial/awesome-rust/tree/ccf576800c1fafffae81933e9d51b71491424f67#build-system) because similar tools (e.g. [kbknapp/cargo-outdated](https://github.com/kbknapp/cargo-outdated)) are listed there too. Happy to put it anywhere else if that makes more sense.

I followed the format specified in the [Contributing Guidelines](https://github.com/rust-unofficial/awesome-rust/blob/ccf576800c1fafffae81933e9d51b71491424f67/CONTRIBUTING.md) even though most entries in this section don't. Happy to adjust if that's preferred.


palfrey:
> I put it in the ["Build system" section](https://github.com/rust-unofficial/awesome-rust/tree/ccf576800c1fafffae81933e9d51b71491424f67#build-system) because similar tools (e.g. [kbknapp/cargo-outdated](https://github.com/kbknapp/cargo-outdated)) are listed there too. Happy to put it anywhere else if that makes more sense.

LGTM
 
> I followed the format specified in the [Contributing Guidelines](https://github.com/rust-unofficial/awesome-rust/blob/ccf576800c1fafffae81933e9d51b71491424f67/CONTRIBUTING.md) even though most entries in this section don't. Happy to adjust if that's preferred.

Yeah, that's fine. On the "we should do this at some point but it's hard" list is "unify all the weird line formats", so the contributing formats are more guidelines than anything else TBH.

PR #1483, opened 2023-04-17
add ockam to libraries/distributed

SanjoDeundiak:
Adding ockam to the list. The distributed category looks a little bit different than others, what do you think about the structure of my addition?


palfrey:
I think the location is good enough. We need at some point to do some wider work regarding reformatting everything anyways.

PR #1485, opened 2023-04-23
Add nutype to README

greyblake:


greyblake:
The build seems to fail due to issue with `atomicdata-dev/atomic-data-rust`, but it looks like it has nothing to do with the change in this PR.

palfrey:
> The build seems to fail due to issue with `atomicdata-dev/atomic-data-rust`, but it looks like it has nothing to do with the change in this PR.

Fixed by e5e968ca6be2e248a750f02803da742413969682

PR #1490, opened 2023-05-10
Added intelli-shell

lasantosr:
[intelli-shell](https://github.com/lasantosr/intelli-shell) is a cli tool to bookmark commands with placeholders support (or fetch them from tldr) and then allow you to search or autocomplete at any time with `ctrl+space`.

It's very useful to avoid having to re-type long commands and to store rarely used commands which tend to be forgotten.

The PR is open as Draft until 50 stars are reached on the repo, which is pretty new.


lasantosr:
Just reached 50 stars, it can be now merged

PR #1500, opened 2023-06-08
Remove WARP terminal from Utilities

arunura:
* WARP is NOT open source.
 * WARP requires user email to sign up on their website to use the app. 
 * WARP sends a ton of marketing emails.
 * This is a strictly commercial product with github used for issue reporting only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,7 +462,6 @@ See also [A comparison of operating systems written in Rust](https://github.com/
 * [tversteeg/emplace](https://github.com/tversteeg/emplace) — Synchronize installed packages on multiple machines
 * [vamolessa/verco](https://github.com/vamolessa/verco) [[verco](https://crates.io/crates/verco)] — A simple Git/Hg tui client focused on keyboard shortcuts
 * [vaultwarden](https://github.com/dani-garcia/vaultwarden#readme) [![Build](https://github.com/dani-garcia/vaultwarden/actions/workflows/build.yml/badge.svg)](https://github.com/dani-garcia/vaultwarden/actions/workflows/build.yml) — Alternative implementation of the Bitwarden server API written in Rust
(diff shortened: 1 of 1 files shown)


palfrey:
None of those are reasons to remove it. There's various other non-open source works here, they've just all fulfilled the two criteria a) being written at least mostly in Rust and b) being "awesome", defined here as enough Github stars or similar as I really don't want to be in the business of going "uh, do I like this?" for entries.

I've added 94ae26b433d6c61bc14018b62431d2701469725d to mark Warp explicitly as commercial software, but otherwise it can stay there.

arunura:
That's a fair conclusion, based on the criteria cited. Thanks for the explanation.

burdges:
Does `:heavy_dollar_sign:` simply mean "source unavailable" like this one?  Or chooses some grey area between what is and is not "open source"?  Or not decided yet? 

If a project has no source available, does anyone really care in what language its written?

arunura:
@burdges said it quite succinctly with
>  I also wonder if a project has no source available then does anyone really care in what language its written?

In my opinion, if the list is about evangelizing Rust in the general software **user** community, it makes sense to included closed source software to say "look at all the cool stuff that can be done with rust". 

But if the list is trying to be a resource for rust developers, not sure how useful it is to have closed source listings in it. 

A reasonable middle ground can be to move all closed source software under an explicit "Closed source" section and add a `:heavy_dollar_sign:` against a listing, if they are commercial as well.

burdges:
I first thought that too, but afaik there is no reason for gradation among the closed vs open source spectrum, so either a section or the sign, but not both.  I only asked because some line still exists, well unless you take the simplest rule:  the source must be available to be in the list, but license is ignored entirely.

PR #1520, opened 2023-07-01
Fix badge links

orhun:
Hello,

I have a few tools on this list which have the wrong badge link.

See: https://github.com/badges/shields/issues/8671

This PR simply updates the links of the badges.


palfrey:
Thanks!

PR #1529, opened 2023-07-17
Add MapLibre/Martin map tile server

nyurik:


nyurik:
It errors out on https://maplibre.org/martin/ -- not certain why - the link is valid

palfrey:
> It errors out on https://maplibre.org/martin/ -- not certain why - the link is valid

You've listed it without the ending slash, so there's a 301. Some 301s are "we're no over here, but this will go away eventually" so we treat it as an error.

nyurik:
oops, thx @palfrey , fixed

PR #1530, opened 2023-07-17
add shoes

cfal:
Add [shoes](https://github.com/cfal/shoes) (a Rust proxy server) to the list


palfrey:
> although it passes, the build action seems to hit many 403 forbidden's during the run phase when accessing the github API

Those are fine. We hit rate-limits when trying PRs (which we don't give auth tokens to), but those are all repos we've previously checked otherwise the build would fail.

PR #1541, opened 2023-08-03
Add Skyfolder

mu-arch:
Added my project Skyfolder (https://github.com/mu-arch/skyfolder) to the Web Servers category.

It has over 50 stars.


palfrey:
This has sorting issues that need fixing (see build log)

mu-arch (commented):
I altered the ordering to be alphabetical

PR #1549, opened 2023-08-10
Update README.md to add the eBPF section

r21gh:
I have included the eBPF section to the libraries in this commit. It comprises three distinct Rust libraries, which could be helpful for the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ If you want to contribute, please read [this](CONTRIBUTING.md).
   * [Date and time](#date-and-time)
   * [Distributed systems](#distributed-systems)
   * [Domain driven design](#domain-driven-design)
-  * [EBPF](#ebpf)
+  * [eBPF](#ebpf)
   * [Email](#email)
   * [Encoding](#encoding)
   * [Filesystem](#filesystem)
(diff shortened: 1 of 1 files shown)


palfrey:
So, as noted, redbpf is archived. It may be mentioned in prominent books, but I'm seeing issues like https://github.com/foniod/redbpf/issues/367 and noting this is folks using a tutorial, and now I'm wondering if the book code works at all. Given the archived status, this isn't getting better there anytime soon, unless there's a fork and a new maintainer.

README.md
```diff
@@ -1191,6 +1192,11 @@ See also [About Rust’s Machine Learning Community](https://medium.com/@autumn_
 
   * [serverlesstechnology/cqrs](https://github.com/serverlesstechnology/cqrs) [[cqrs-es](https://crates.io/crates/cqrs-es)] — A framework for CQRS and event sourcing with [user guide](https://doc.rust-cqrs.org/)
 
+### EBPF
```
palfrey, line 1195: Why the capitalisation? Shouldn't this be "eBPF"?

r21gh: @palfrey eBPF is an unequivocally recognized acronym. Would the linter approve of a sentence that begins with a lowercase letter?

palfrey: I believe so. It's mostly checking "does the top-level ToC equal the actual sections".

PR #1558, opened 2023-08-28
Add Rust video series playlist 🦀

pcgeek86:


palfrey:
Thanks for the vim fix, but I'm going to reject this right now. So, our standards on Github repos is 50 stars or 2000 crates downloads. We don't have a standard for Youtube videos, but given that the most views any of those videos has is 235 right now, that feels well below an equivalent standard for "the community likes this" currently and as the [CONTRIBUTING.md](https://github.com/rust-unofficial/awesome-rust/blob/main/CONTRIBUTING.md) notes, we're using metrics like this to try and be at least vaguely objective.

I will note that we haven't necessarily been very consistent with this in some of the other Youtube links, but that's mostly due to a combination of a lack of automated tooling so far and maintainer burnout, and that will probably get fixed at some point.

If you can get the views for any of the videos up to a few thousand at least, then please re-open this.

PR #1576, opened 2023-10-10
Update README.md

rodrigorc:
Add my Papercraft program to the Graphics section.

Properly it should be into an "Arts&Crafts" section, but that doesn't currently exist, and I don't know if it is worth it.


palfrey:
> Properly it should be into an "Arts&Crafts" section, but that doesn't currently exist, and I don't know if it is worth it.

Probably not yet, but will try and keep an eye on that.

PR #1583, opened 2023-10-18
Add Rinf: Rust in Flutter

temeddix:
This PR adds [Rinf](https://github.com/cunarist/rinf) framework that connects Flutter and Rust


temeddix:
Hmm, there seems to be some CI error that cannot be fixed on my side

PR #1592, opened 2023-11-02
Add Jon Gjengset's stream

VENCYR:


README.md
```diff
@@ -1827,6 +1827,7 @@ A registry allows you to publish your Rust libraries as crate packages, to share
 * [RustBooks](https://github.com/sger/RustBooks) — list of RustBooks
 * [RustCamp 2015 Talks](https://www.youtube.com/playlist?list=PLE7tQUdRKcybdIw61JpCoo89i4pWU5f_t)
 * [RustViz](https://github.com/rustviz/rustviz) — generates visualizations from simple Rust programs to assist users in better understanding the Rust Lifetime and Borrowing mechanism.
+* [Watch Jon Gjengset Implement BitTorrent in Rust](https://www.youtube.com/watch?v=fhEHBQog-Cc&t=6606s&ab_channel=JonGjengset)
```
palfrey, line 1830: This appears to be a link to almost the end of the stream? Also it's an unlisted video - a) should we have the link posted here and b) where did it come from, as that might be the better link.

VENCYR: Thanks! Can we use this link instead please? - https://www.youtube.com/watch?v=jf_ddGnum_4&t=176s

palfrey: That at least solves the unlisted video problem! If you strip the `&t=` parameter entirely it'll be the beginning of the video. Can you please edit your PR to use that new URL?

VENCYR: Updated!

PR #1596, opened 2023-11-05
Add jetbrains RustRover

mxsm:
- Add jetbrains RustRover IDE


palfrey:
Firstly, this isn't part of Intellij AFAIK. I mean, it's another Jetbrains IDE, but that's not quite the same. Secondly, do you have any publicly available metrics for this one? Open source bits we can easily do the stars, but commercial works need to prove notability. Yes, it's Jetbrains, but that doesn't get it automatic "we'll just add it".

PR #1602, opened 2023-11-13
Add Rauthy OIDC IAM

sebadob:
This adds [Rauthy](https://github.com/sebadob/rauthy) to the README.

OpenID Connect Single Sign-On Identity & Access Management  written in Rust.


sebadob:
I don't get it where this 403 in the build pipeline might come from. Is this some random thing that just happens sometimes? The links its complaining about are working fine for me.

palfrey:
> I don't get it where this 403 in the build pipeline might come from. Is this some random thing that just happens sometimes? The links its complaining about are working fine for me.

Basically Github gets grumpy when you try and check this many repos. I've fixed that by deleting the cache which meant it didn't need to check as many in one go.

sebadob:
Ahh got it, thank you!

PR #1616, opened 2023-12-05
adding broot, yazi, joshuto and xplr

TornaxO7:


palfrey:
This needs updating to current main, but also the ToC lint is failing because you haven't added it to the top-level ToC. Why the two levels of ToC added as well? Why not just "File Managers", and we can split out to Terminal if we get many non-Terminal ones.

TornaxO7:
@palfrey done :) Thank you pointing out the issue with the TOC :)

PR #1617, opened 2023-12-06
Add Chess-tui to games

thomas-mauran:
- add chess-tui to games

### Demo
demo


palfrey:
Sort ordering needs fixing. See build log.

thomas-mauran:
> Sort ordering needs fixing. See build log.

fixed

PR #1625, opened 2023-12-27
Add Festival

hinto-janai:
https://github.com/hinto-janai/festival


palfrey:
Can you update your branch to current `main` and then the build should work. Thanks!

hinto-janai:
@palfrey done :)

PR #1651, opened 2024-01-29
feat: add async-dropper to asynchronous libs

t3hmrman:
Thanks for the awesome listing! I added my project `async-dropper` which is for people who want an `AsyncDrop` -- it has two implementations (one derive macro based, the other a simple wrapper object)


palfrey:
See build logs for some stuff that needs correcting.

t3hmrman:
Ahh @palfrey thanks -- looks like I have some spaces and maybe emacs chopped the newline from the end of the file, will fix

t3hmrman:
Should be fixed, and rebased to master now

palfrey:
Sort order also needs fixing. See logs.

t3hmrman:
Thanks for the corrections!

PR #1653, opened 2024-01-29
Added mirrord

emanuelef:
Useful tool to run and debug locally Kubernetes pods.
Not sure about the category though.


palfrey:
> Not sure about the category though.

I think it's good enough given we've got linkerd there as well, but could do with some reorganisation at some point.

PR #1672, opened 2024-03-05
Cleanup README

joshka:
Use consistent em-dashes, and spacing

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,14 +126,14 @@ See also [Rust — Production](https://www.rust-lang.org/production) organizatio
 * [doprz/dipc](https://github.com/doprz/dipc) — Convert your favorite images and wallpapers with your favorite color palettes/themes [![crates.io](https://img.shields.io/crates/v/dipc)](https://crates.io/crates/dipc)
 * [Factotum](https://github.com/snowplow/factotum) — A system to programmatically run data pipelines
 * [fcsonline/drill](https://github.com/fcsonline/drill) — A HTTP load testing application inspired by Ansible syntax
-* [fend](https://github.com/printfn/fend) - Arbitrary-precision unit-aware calculator [![build](https://github.com/printfn/fend/workflows/build/badge.svg)](https://github.com/printfn/fend)
+* [fend](https://github.com/printfn/fend) — Arbitrary-precision unit-aware calculator [![build](https://github.com/printfn/fend/workflows/build/badge.svg)](https://github.com/printfn/fend)
 * [Fractalide](https://github.com/fractalide/fractalide) — Simple microservices
 * [habitat](https://github.com/habitat-sh/habitat) — A tool created by Chef to build, deploy, and manage applications.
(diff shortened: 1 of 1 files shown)


palfrey:
In lots of ways, I like this. I like improvements to the consistency of the list, especially ones that basically take stuff that was being kinda done and move it to "and now it all does it".

My concern however is that maintenance on this will be a pain in the proverbial. Pretty much everything else that's being maintained quality wise right now on the repo is automated. This is very deliberate as it means I have a hope in heck of keeping up, just by being able to go "if the CI is happy, merging won't be that bad". This PR if merged would make a bunch of improvements, and then probably the improvements would go away over time and I just wouldn't notice.

If you can figure out a way to automate this (ideally by improvements to the existing Rust tool, but I'll take whatever), I'd be interested in merging this.

joshka:
Once the consistency is there automating the lint will be fairly simple (each line matches a regex). 

The rationale behind this change was that there is a tool that uses this data for visualization and a bunch of lines don't match the regex in the tool.

joshka:
Got it - I'll add this at some point soon

palfrey:
> Got it - I'll add this at some point soon

How's this going?

joshka:
> How's this going?

Done - I think
I used the linter to find all the places in the readme that didn't have descriptions, and added them (mostly copied from the GitHub description / crate description verbatim, but some copied from a readme or similar)

Merged from the main branch into this branch - I'd probably suggest squash merging this rather than hitting the merge button for a neater history.

src/main.rs
```diff
@@ -578,6 +580,9 @@ async fn main() -> Result<(), Error> {
                                     warn!("No valid crates link");
                                 }
                                 return Err(format_err!("Not high enough metrics ({:?} stars < {}, and {:?} cargo downloads < {}): {}", github_stars, required_stars, cargo_downloads, MINIMUM_CARGO_DOWNLOADS, list_item));
+                                if !ITEM_REGEX.is_match(&list_item) {
+                                    return Err(format_err!("Item does not match the template: {}. See https://github.com/rust-unofficial/awesome-rust/blob/main/CONTRIBUTING.md#tldr", list_item));
+                                }
```
palfrey, line 585: This is code after a `return` which the compiler is flagging as "unreachable expression"

joshka: Indeed it is. My apologies. I was testing this with some of the code commented out so as not to trigger the throttling issues, and uncommented things wrong at the last minute before pushing. Will fix and retest.

src/bin/hacktoberfest.rs
```diff
@@ -171,7 +171,7 @@ async fn main() -> Result<(), Error> {
             return;
         }
         used.insert(url.clone());
-        if results.get(&url).is_some() {
+        if results.contains_key(&url) {
```
joshka, line 174: This is picked up by the clippy deny rule added recently.

src/main.rs
joshka: This is fixed

palfrey:
Awesome. Thanks for this!

PR #1684, opened 2024-03-17
Actions upgrade for node 16 issues

palfrey:
Fixes all bar the always-upload-cache. I've filed a bug there (https://github.com/pat-s/always-upload-cache/issues/3) to fix it, or at some point `actions/cache` might be usable for this now, but that has it's own issue (https://github.com/actions/cache/issues/1315)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,9 +21,8 @@ jobs:
           components: rustfmt
       - uses: Swatinem/rust-cache@v2
       - name: Load results cache
-        uses: actions/cache@v4
+        uses: pat-s/always-upload-cache@v3.0.11
         with:
-          save-always: true
           path: results/*.yaml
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -14,19 +14,18 @@ jobs:
...
           components: rustfmt
       - uses: Swatinem/rust-cache@v2
-      - name: Get random cache id
-        run: echo "CACHE_ID=$((RANDOM))" >> $GITHUB_ENV
-        shell: bash
-      - uses: pat-s/always-upload-cache@v3.0.11
+      - name: Load rust cache
+        uses: actions/cache@v4
         with:
+          save-always: true
```
palfrey, line 26: Fun facts, this doesn't actually work. See https://github.com/actions/cache/issues/1315

PR #1714, opened 2024-04-24
Update README.md - Adding sample code showing enviornmental sensing features like sensing changes in water tanks level

joeatbayes:
Adding examples for environmental sensing


joeatbayes:
Realized I put in relative URI rather than absolute

PR #1715, opened 2024-04-24
Update README.md - Added Section for enviornmental Sensing

joeatbayes:
Added section for Environmental Sensing along with links to sample code for two different ways of reading variable capacitors like water level in tanks.  Also included a couple links showing how to interact with RS485 over 1/2 duplex chips More to come in that area.    We may want to consider broadening this section to Environmental Sensing & Process Control.  

I included the main topics shown for each example because I have found it difficult to locate Rust code that contains specific features I need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1889,6 +1889,49 @@ A registry allows you to publish your Rust libraries as crate packages, to share
 * [RustViz](https://github.com/rustviz/rustviz) — generates visualizations from simple Rust programs to assist users in better understanding the Rust Lifetime and Borrowing mechanism.
 * [Watch Jon Gjengset Implement BitTorrent in Rust](https://www.youtube.com/watch?v=jf_ddGnum_4)
 
+
+## Enviornmental Sensing
+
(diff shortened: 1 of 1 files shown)


palfrey:
In order to attempt to be as fair as we can with the process regarding additions of new projects we're using external popularity metrics to guide this. Our current standards are at least one of: a minimum of 50 stars on Github; 2000 downloads for projects with published crates; or equivalent values in other areas (e.g. downloads in other places, star-type popularity in other repositories).

joeatbayes:
To reach the 50 starts requires that people find the work.  I don't have time to go do a bunch of marketing just to share work that  could be of value trying to get started with rust on ESP32 to do real work.   If you don't want to help those people find a valuable resource then it is unfortunate for your users.

palfrey:
https://github.com/rust-unofficial/awesome-rust/blob/main/CONTRIBUTING.md is very clear on why we don't want to do that.

PR #1716, opened 2024-04-24
Because they should only slightly diff from main

palfrey:
And without this you end up with outdated caches and we only really fully update on main (because that's got the creds to do a full update)


palfrey:
Blocked on either https://github.com/pat-s/always-upload-cache/issues/4 or https://github.com/actions/cache/issues/1315

PR #1740, opened 2024-06-01
Update `icepuma/rust-action` to `Mirlahiji/rust-action`

ranger-ross:
I noticed that [icepuma/rust-action](https://github.com/icepuma/rust-action) is no longer maintained and the README suggested using [Mirlahiji/rust-action](https://github.com/Mirlahiji/rust-action) (a fork) instead.


palfrey:
So the tricky bit here is the new repo is at only 9 stars, which is below our thresholds. I'm also not seeing any other public metrics that indicate it's more popular than that? In the absence of such matters, I'd either want to keep the old repo around and see if the new one gets more popular, or just delete it entirely TBH.

ranger-ross:
Thats fair enough :+1: 
Sorry, I didn't realize there was a star requirement until after I had already opened the PR :sweat_smile:

PR #1755, opened 2024-06-27
Use the easier to type dash for name separation

palfrey:
The other dash is a nice idea, but it's a) a pain to find on keyboards and b) everyone submitting stuff tends to type the other one, and explaining things is painful so swapping back.


joshka:
Makes sense - when I fixed that regex I went with the more common rather than easier form. I prefer the easier dash like you've mentioned and probably should have called this out.

PR #1758, opened 2024-07-08
Added cel-rust, common expression language interpreter

clarkmcc:


palfrey:
I think you're hitting the problem that a while back I did https://github.com/rust-unofficial/awesome-rust/pull/1755 which reset the dash between repo and description to a standard `-` not the longer one I actually can't even find on my keyboard, but didn't fix it in the docs. https://github.com/rust-unofficial/awesome-rust/commit/98645308c81b6a17a130ca862927db06618a20b7 fixes that in CONTRIBUTING.md and at some point I need to rewrite the regex matcher to have better diagnostics. Sorry about that!

clarkmcc:
Huh, didn't realize there were two different kinds... Thanks!

PR #1760, opened 2024-07-13
Add Rudis

sleeprite:
https://github.com/sleeprite/rudis


palfrey:
Sort ordering needs fixing (see build logs)

sleeprite:
> Sort ordering needs fixing (see build logs)

I have adjusted the sorting order according to the rules

PR #1762, opened 2024-07-14
Removed crates - repo archived

Squidroot2:
Found this when I ran the build locally without the results yaml files. It failed because archived repos are treated as zero stars.

The https://github.com/serayuzgur/crates repo has been archived and is now pointing users to https://github.com/filllabs/dependi (a typescript project so no replacement for this list)


palfrey:
Thanks!

PR #1763, opened 2024-07-16
Replaced "Better TOML" (deprecated) with "Even Better TOML"

0xdea:
Replaced "Better TOML" VSCode extension (deprecated) with "Even Better TOML" (currently in preview).


palfrey:
You'll also need to replace https://github.com/rust-unofficial/awesome-rust/blob/main/src/main.rs#L56

0xdea:
Thanks, I has missed this!

PR #1764, opened 2024-07-18
add clipboard-rs to lib-system

ChurchTao:
https://github.com/rust-unofficial/awesome-rust/pull/1656

Now is 6000+ downloads.


palfrey:
Doesn't have a crates link, and we don't have a magic way to get those from a github link, so will need explicitly adding.

ChurchTao:
> Collaborator

Sorry, I re-added the ci and crates links

palfrey:
Build should work again once https://github.com/rust-lang/arewewebyet/issues/439 is fixed

PR #1766, opened 2024-07-24
Add ratchet_rs

SirCipher:


palfrey:
> @palfrey, unless I'm misreading the logs I don't think that error is related to what I've changed?

No, but also not quite sure why it's failing like that. Let me poke some stuff...

SirCipher:
@palfrey, thank you

PR #1772, opened 2024-07-28
Remove pikkr

equal-l2:
No update since 7 years ago and it doesn't compile on the latest stable.


equal-l2:
The error seems pre-existing.
(`https://github.com/subspace/subspace` is moved to `https://github.com/autonomys/subspace`

palfrey:
Seems to compile fine for me (1.80.0). It's got a warning about if it got updated to 2021 edition, but you can happily mix editions, so I'm not seeing the issue here? Even if it's not touched for 7 years, if it's not actively totally broken or archived, it can still be useful.

equal-l2:
The build error is not observed on the master, but the latest version on crates.io causes lots of build errors (note that it's NOT a warning, meaning the package is unusable).
I do not insist that the package must be deleted from the list, but IMO every package on the list should be usable without extra caveat.

palfrey:
Ah, yup. Poked at stuff, filed https://github.com/pikkr/pikkr/issues/76 and will merge this.

palfrey:
@equal-l2 There's conflicts with this now, probably with https://github.com/rust-unofficial/awesome-rust/pull/1773. Can you resolve those please?

PR #1773, opened 2024-07-28
Replace json-rust with jzon-rust

equal-l2:
`json-rust` is now regarded as abandoned.
https://rustsec.org/advisories/RUSTSEC-2022-0081.html

`jzon-rust` is a community-driven fork of `json-rust`.


palfrey:
You'll need to fix the sort ordering (see build log)

PR #1774, opened 2024-07-28
Add sonic-rs

equal-l2:
`sonic-rs` is a very fast JSON library based on SIMD.


palfrey:
You'll need to fix the sort ordering (see build log)

equal-l2:
@palfrey Thank you for the help! Fixed and rebased onto main.
Would you check again, please?

PR #1778, opened 2024-08-03
Update README.md

deadsoul44:
perpetual-ml/perpetual — A self-generalizing gradient boosting machine which doesn't need hyperparameter optimization.


deadsoul44 (commented):
changes to comply with the template

PR #1787, opened 2024-08-14
add `Peroxide` numeric library

RobWalt:
I was looking for something like this and the options listed are not really helpful since:

- scirust is unmaintained since 7 year. (Should we just scratch it at this point?)
- russell isn't written in pure rust which makes it less cross platform than `Peroxide` and I need that
- `peroxide` is actively maintained

### DIFF
--- a/README.md
+++ b/README.md
@@ -1096,6 +1096,7 @@ See also [About Rust’s Machine Learning Community](https://medium.com/@autumn_
 * Science
   * [cpmech/russell](https://github.com/cpmech/russell) - Rust Scientific Library for numerical mathematics, ordinary differential equations, special math functions, high-performance (sparse) linear algebra
   * [indigits/scirust](https://github.com/indigits/scirust) - Scientific computing library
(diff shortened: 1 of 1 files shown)


palfrey:
>     * scirust is unmaintained since 7 year. (Should we just scratch it at this point?)

Can you remove this please? The long non-maintenance is less of a concern, but the fact that it's most recent release [doesn't compile](https://docs.rs/crate/scirust/0.0.5/builds/103750) and the repo appears to have undergone radical changes since then but no release is going to make it hard to use.

RobWalt:
> > ```
> > * scirust is unmaintained since 7 year. (Should we just scratch it at this point?)
> > ```
> 
> Can you remove this please? The long non-maintenance is less of a concern, but the fact that it's most recent release [doesn't compile](https://docs.rs/crate/scirust/0.0.5/builds/103750) and the repo appears to have undergone radical changes since then but no release is going to make it hard to use.

Done 👍🏼

PR #1800, opened 2024-09-04
add spring-rs

holmofy:
spring-rs is a application framework written in rust inspired by java's spring-boot


palfrey:
Sort ordering needs fixing, see build logs

PR #1807, opened 2024-09-17
Add Halloy link

sharma-shray:
Hi, its client for Mac, Windows, and Linux platforms.


README.md
```diff
@@ -1677,6 +1677,9 @@ See also [Are we game yet?](https://arewegameyet.rs)
   * [eclipse-zenoh/zenoh](https://github.com/eclipse-zenoh/zenoh) - Zero Overhead Network Protocol
 * ZeroMQ
   * [erickt/rust-zmq](https://github.com/erickt/rust-zmq) - [ZeroMQ](https://zeromq.org/) bindings
+* Halloy
```
palfrey, line 1680: Halloy isn't a protocol, so this doesn't belong here. Theoretically we could do this as `IRC` as the protocol, but it's an application that does IRC, not an IRC library, so it should go under `Applications`.

palfrey:
Closed awaiting responses

PR #1809, opened 2024-09-27
add brush shell

reubeno:
I'm proposing adding [brush](https://github.com/reubeno/brush) to system tools list, now that its download count on [crates.io](https://crates.io/crates/brush-shell) has gone above 3,000:

[Crates.io](https://crates.io/crates/brush-shell)

I'm the primary maintainer of the project and happy to answer any questions about it. This is also my first proposed addition to this list and am happy to take any feedback you have on this change. Thanks in advance for your consideration!


reubeno:
I was missing a badge for the crates.io page for this project; I've since added that.

PR #1817, opened 2024-10-06
Add zipxing/rust_pixel in games

zipxing:
https://github.com/zipxing/rust_pixel


palfrey:
Sorting needs fixing, see build logs

PR #1823, opened 2024-10-10
Add samba

koppor:
This adds [samba](https://gitlab.com/ballroom-dancing/samba). I like this software, because it combines Rust and VueJS

In a nutshell, samba aims to be of service (and not in the way) of dancing clubs and active dancers, for regular practice, competition practice, and all the other dancing related matters.

Library view

Edit practice session


palfrey:
The build fails mostly because it can't check Gitlab for automatic metric requirements (as per our [CONTRIBUTING.md](https://github.com/rust-unofficial/awesome-rust/blob/main/CONTRIBUTING.md)). OTOH, I'm also only seeing 6 stars there, so it currently fails our popularity metrics. Is there any other source of metrics I should be looking at?

koppor:
> The build fails mostly because it can't check Gitlab for automatic metric requirements (as per our [CONTRIBUTING.md](https://github.com/rust-unofficial/awesome-rust/blob/main/CONTRIBUTING.md?rgh-link-date=2024-10-10T21%3A26%3A04Z)). 

Oh, I missed that. Sorry for that!

PR #1839, opened 2024-11-14
Add "binarly-io/idalib" under the new "reverse engineering" category

0xdea:
I'm using this library to develop reverse engineering and security tools. I think it would be a great addition to the list, but I couldn't find a suitable category, so I added one. I hope it's ok!


0xdea:
There seems to be a problem in the build workflow that's unrelated to my changes...

PR #1866, opened 2024-12-28
Add "Finance" app category and Tackler-NG accounting application

35VLG84:
Hello, 

This PR adds similar "Finance" application category as there is for libraries.  

And secondly, it adds Tackler-NG to that Finance  application category.  Tackler-NG is fast, 
reliable bookkeeping engine with native GIT SCM support for plain text accounting.

Github: https://github.com/e257-fi/tackler-ng
Crates.io: https://crates.io/crates/tackler

Sorry to stepping ahead if the  finance application category doesn't make sense, then please just reject this PR and I will make a new one without it. In that case, would the top level Application list be ok for tackler?

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,7 @@ See also [crates matching keyword 'emulator'](https://crates.io/keywords/emulato
 
 See also [Payments](#payments) applications.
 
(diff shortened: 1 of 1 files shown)


palfrey:
LGTM

PR #1870, opened 2025-01-03
Add cargo-testdox

bitfield:
Thanks for your hard work maintaining this amazing repo!


bitfield:
I think I'm missing something here—the contributing guide says:

> if you want to add something, please use the template `[ACCOUNT/REPO](https://github.com/ACCOUNT/REPO) [[CRATE](https://crates.io/crates/CRATE)] — DESCRIPTION`

That's the template I used, complete with hyphen, but it's rejected by the CI check:

> "bitfield/cargo-testdox [cargo-testdox] — Turns your Rust tests into docs CI" uses a '—' hyphen, not the '-' hyphen and we enforce the use of the latter one

Could you help me understand the mismatch, please?

PR #1904, opened 2025-03-11
docs: added a link to the lint workflow

operagxoksana:


palfrey:
You'll need to add alt-text to the new image to pass the build.

operagxoksana:
you mean this link
https://github.com/rust-unofficial/awesome-rust/actions

operagxoksana:
Should I add this icon?

palfrey:
You don't need an icon, it needs alt-text. You added `[![]` and the `[]` should contain a piece of text that represents the image for cases where either the image doesn't load, or someone's using a screen reader. "lint badge" would do fine.

PR #1905, opened 2025-03-12
Update 'Awesome Rust Books' link

bitfield:
The old repo appears to be abandoned; this fork is more up to date and is actively maintained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2001,6 +2001,7 @@ A registry allows you to publish your Rust libraries as crate packages, to share
 * Learning
   * [100 Exercises To Learn Rust](https://rust-exercises.com) - Learn Rust through 100 hands-on exercises, covering syntax, types, and more
   * [Aquascope](https://github.com/cognitive-engineering-lab/aquascope) - Interactive visualizations of Rust at compile-time and run-time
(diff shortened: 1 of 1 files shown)


palfrey:
The original repo hasn't had updates for 4 months, but I'm not seeing any issues going something like "hey, is this still maintained", or some sort of "this is archived" on the repo, so I'm not calling it dead yet.

Additionally, it's got 4.8k stars v.s. the 11 of your proposed repo so by our usual standards your fork wouldn't get approved here. Documentation links aren't getting the same enforcement of stars as other links because a lot of them are non-Github so it wasn't as useful, but I'm not seeing a high enough threshold here yet. You do have the most starred of the forks so far, but it's still well below our usual 50 star threshold.

I'm closing this for the moment. If the situation changes with the original repo, please reopen.

bitfield:
Roger that @palfrey, thanks for the suggestion: I've opened https://github.com/sger/RustBooks/issues/84 to check.

It may take a while for my fork to catch up to the old one's collection of stars, indeed, but as you say perhaps that isn't the most relevant metric here. My collection has more, and more recent, Rust books, and I would guess is thus likely to be the one that Rust learners would find more useful.  But it's only 24 hours old, so let's see if it picks up a few more stars over time.

PR #1907, opened 2025-03-14
added libs.tech

sizovs:
Added libs.tech (direct link to Rust category). Libs.tech helps devs stay on top of ecosystems they're interested in.

https://libs.tech/rust


palfrey:
So, I'm not quite seeing the benefit here? This feels like a) a competitor to this list and b) not one with any public popularity metrics I'm seeing to indicate this project should get added.

palfrey:
Closing due to lack of response

PR #1955, opened 2025-04-27
Delete dropped rust production

hg127:
rust-lang/www.rust-lang.org#2119 dropped it


palfrey:
Done in https://github.com/rust-unofficial/awesome-rust/pull/1956 which fixed some other issues

PR #2038, opened 2025-06-18
Added Kaspa

KranzAklilu:


burdges:
Proof-of-work is never fast.  That's kinda the point.

"Scalable" always means data partitioning, horizontal and/or vertical.  A single blockchain is never scalable because its like a single "table" in a single database.  

If you want the long version..

A "(decentralized) scalable blockchain" means some distributed sysrtem where nodes cannot access all the data, or verify every block, but nevertheless some argument proves security under cryptographic and byzantine-like assumptions (or at least some social argument keeps away attackers).

A rough quick list of the 6-ish known "(decentralized) scalable blockchain" protocols: 
1. Bridges suffice if all blockchains have identical validator sets.  You cannot ask small operators to add nodes whenever, but this works when validators are large organizations, like say governments signatory to some treaty.  If validator sets disagree like cosmos then scaling degrades security. 
2. True (advance) sharding of validator sets uses [concentration inequality arguments](https://bford.info/pub/dec/omniledger-abs/)
3. Randomized auditing can be made provably secure too, ala polkadot's [ELVES](https://eprint.iacr.org/2024/961).

None of 1-3 would ever support proof-of-work, but if one adds latency between shards, then several other scalability protocols do support proof-of-work:

4. SNARKs could prove correctness, ala zk roll ups, but incur high cross shard latency and eenergy costs from provers.
5. "Optimisitc" roll ups lack the formal security arguments of 2 or 3, but instead they impose very high cross shard latency (one week usually).
6. State channels like lightning network exploit application specific security arguments and data locality, so perfect for some applications like few player games, but quite limited or centralizing for typical financial applications.  They incur latency under aborts.

These 6-ish protocols only enable data partitioning within the distributed system, but the actual data partitioning is inherently applications specific, just like in web2, aka general smart contracts can never be scalable, and the data partitioning should respect the cross shard latency.

KranzAklilu:
->"Proof of work is never fast. That's kinda the point." is a misleading statement. Obviously, miners need to solve the block inorder to confirm. The difficulty varies based on the hashrate. When we say a network is fast, we never mean it's easy to solve a block.
Unlike bitcoin producing 1 block every 10 mins, Kaspa network generates 10 blocks every second, there's no other proof of work that gets close to that. Also, transactions get fully confirmed in 1 second (given this points, one could even say it's the easiest to solve).
-> "Scalable" always means data partitioning, horizontal and/or vertical. A single blockchain is never scalable because its like a single "table" in a single database.
It doesn't. Scalability is the property of a system to handle a growing amount of work (from Wikipedia). Data partitioning is a technique used to achieve scalability. In Kaspa given that average rate of 10 block every second allows it to pack in a large number of transactions in very short duration of time and is able to meet any load.
Also for your point about a single blockchain being like a single table, Kaspa utilizes blockDag design which allows simultaneous block creation.
Thanks for your comment. Let me know if you think I'm mistaken in any of my points.
At the end of the day, no one can argue it's an awesome-rust project 😉

palfrey:
My personal opinion is that all the blockchain everything should be removed, as they're all not even slightly awesome and actively causing problems in the world.

However, this is why the criteria in the contributing file is stars and similar levels of "does the community like this", mainly so us maintainers do not have to make judgement calls like that.

PR #2058, opened 2025-07-13
Add topheman/snake-pipe-rust to Game section

topheman:
Added https://github.com/topheman/snake-pipe-rust to the game section - A snake game in the terminal based on stdin/stdout (+tcp and unix domain sockets).


palfrey:
The repo fails the stars requirement but there's enough crate downloads. If you add a link to that as well please, as per existing other entries that'll work.

PR #2096, opened 2025-10-05
Add reaction to Security tools

ppom0:
It has "only" 31 stars right now, but it's on a much smaller forge than GitHub, so I guess it means more.

It's the 12th most starred project there: https://framagit.org/explore/projects/starred


palfrey:
Can you add an entry to the bottom of https://github.com/rust-unofficial/awesome-rust/blob/main/src/main.rs#L85C90-L85C147 noting it now has 55 stars, and that'll fix the build issue. Thanks!

ppom0:
Or if I close and re-open the PR, will it pass?

palfrey:
> Or if I close and re-open the PR, will it pass?

No, because the build tooling only supports Github for automatic stars detection. It's a bunch of work to support other platforms, so it's mostly been easier to just use the override functionality as I detailed above.

ppom0:
Done!

PR #2101, opened 2025-10-13
Add Free Ferris Pack

MariaLetta:
Hey guys,

I would like to add link to **Free Ferris Pack** which is my latest work with Ferris illustrations published under CC0 license.
Also added _Arts_ subcategory because I didn't find other appropriate. Not completely sure about _Resources_, but looks like it's relevant. Please let me know if there is anything I need to fix in this commit. 🦀💖✨

Thanks!


palfrey:
Good location, happy with it being in there. Arts should go after `A Brief History of Rust` as per the build logs.

MariaLetta:
Thanks @palfrey should fixed now!

PR #2111, opened 2025-10-23
Add Dash documentation browser

Kapeli:
Added Dash for browsing Rust docs offline


palfrey:
This isn't really a Rust project. It happens to incidentally support Rust, and given the existing support from `cargo doc`, I'd question even that level of need here. If it was open source, then this might be of more interest, but our bar for closed source projects is pretty high.

PR #2195, opened 2026-02-06
add blinc ui framework

alexn-s:
It is a new ui framework i found. It has more than 200 stars.


palfrey:

alexn-s:
Ahh ok. I previously inserted be the project name only. Now the order should be fixed.

PR #2206, opened 2026-02-13
Add sql-splitter to Database section

HelgeSverre:
Adds [sql-splitter](https://github.com/HelgeSverre/sql-splitter) to the **Database** section.

sql-splitter is a CLI toolkit for processing SQL dump files, built in Rust with a streaming architecture for constant memory usage on large files. Features include split, merge, convert, validate, sample, shard, redact, diff, graph, order, and query commands across MySQL, PostgreSQL, SQLite, and MSSQL.

- Uses DuckDB for embedded SQL analytics
- Supports compressed input (.gz, .bz2, .xz, .zst)
- Available on crates.io: `cargo install sql-splitter`

Website: https://www.sql-splitter.dev


palfrey:

HelgeSverre:
Fair enough :) tnx for the heads up

PR #2243, opened 2026-03-01
Add wreq-ruby to FFI / Ruby section

ZilvinasKucinskas:
wreq-ruby is the first production-ready Ruby HTTP client with real browser TLS/HTTP2 fingerprinting — and it's powered by Rust.

Ruby has never had a library capable of emulating browser TLS and HTTP2 fingerprints. Every existing Ruby HTTP library relies on OpenSSL, which provides no control over TLS extension ordering, cipher suite ordering, or HTTP/2 frame settings. Python, Go, and Node have had solutions for this. Thanks to the wreq crate (403k+ downloads), Ruby now does too.

wreq-ruby wraps the wreq crate via FFI with pre-compiled native gems for Linux and macOS, so Ruby developers don't need a Rust toolchain. It's used in production at SearchApi and actively maintained.


palfrey:

PR #2255, opened 2026-03-05
Add sabiql to Database section

riii111:
Add [sabiql](https://github.com/riii111/sabiql) to the Database section.

https://github.com/riii111/sabiql

sabiql is a fast, driver-less TUI to browse, query, and edit PostgreSQL databases, built with Ratatui. It uses `psql` CLI under the hood instead of requiring a database driver.

- Stars: 123 (meets the 50+ threshold)
- Published on crates.io: https://crates.io/crates/sabiql


palfrey:

riii111:
@palfrey 
Thanks for the review! Fixed the sorting order.

PR #2267, opened 2026-03-08
Add RusTorch to Machine learning section

Genius-apple:
I've added RusTorch to the Machine learning section.

RusTorch is a production-ready deep learning framework written in pure Rust. It provides a PyTorch-like API, a complete autograd engine, and neural network modules ( nn , vision , text ), making it easier for Python developers to transition to Rust for ML tasks.

- Repository : https://github.com/Genius-apple/RusTorch
- Crate : https://crates.io/crates/rus-torch
It is actively maintained and documented. Thanks for considering!


palfrey:

PR #2289, opened 2026-03-15
Add SuperLightTUI to TUI libraries

subinium:
Add [SuperLightTUI](https://github.com/subinium/SuperLightTUI) to the TUI libraries section.

SuperLightTUI is an immediate-mode TUI library for Rust with 50+ built-in widgets, flexbox layout, Tailwind-style chaining API, 7 theme presets, animation system (Tween/Spring/Keyframes), and half-block image rendering. Two core dependencies (crossterm + unicode-width), zero unsafe code.

```rust
fn main() -> std::io::Result<()> {
    slt::run(|ui| { ui.text("hello, world"); })
}
```

- [crates.io](https://crates.io/crates/superlighttui)
- [docs.rs](https://docs.rs/superlighttui)
- CI passes ✓


palfrey:

PR #2294, opened 2026-03-17
Add assay — MCP tool-call firewall with evidence trail

Rul1an:
[assay](https://github.com/Rul1an/assay) is a Rust CLI that acts as a firewall for MCP (Model Context Protocol) tool calls.

It intercepts every tool call, evaluates it against a YAML policy, and produces a replayable evidence bundle. Deterministic — same input, same decision, every time.

- Published on crates.io: [assay-cli](https://crates.io/crates/assay-cli)
- CI badge: passing
- MIT licensed
- Covers 7/10 OWASP MCP Top 10 risks

Added to the Security tools section.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -516,6 +516,7 @@ See also [A comparison of operating systems written in Rust](https://github.com/
 
 ### Security tools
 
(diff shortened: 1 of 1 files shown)


palfrey:

PR #2298, opened 2026-03-19
Add Optimization Engine link to README

alphaville:
Optimization Engine (OpEn) is a fast numerical optimization library written in Rust

- Add link to [GitHub](https://github.com/alphaville/optimization-engine) page and [crates.io](https://crates.io/crates/optimization_engine)
- Add GitHub Actions badge

### Description

Optimization Engine (OpEn) is a numerical optimization library written in Rust and a code generator in Python that facilitates the design of optimizers, suitable for embedded applications and robotics. Typical applications include model predictive control (MPC) and moving horizon estimation (MHE), which are popular in robotics. OpEn has been used on ground and aerial vehicles. 

Some examples of applications where OpEn has been used are [autonomous racing cars](https://giuseppesilano.net/publications/SMC22.pdf), navigation of a [Husky robot](https://alphaville.github.io/optimization-engine/docs/example_navigation_ros_codegen) using ROS, collision-free navigation of heavy equipment ([paper](https://arxiv.org/pdf/2404.14257), [demo](https://youtu.be/YrUXZ3_oJlU)).

OpEn can automatically generate ROS packages, which can be used directly in robotics applications. 

Lastly, OpEn is becoming popular: it currently counts 616 stars on GitHub and more than 300k downloads from [crates.io](https://crates.io/crates/optimization_engine). The Python package, `opengen` counted 804 downloads in the last month ([link](https://pypistats.org/packages/opengen)).

### Why should it be included?

The unique features of OpEn are: (i) the solver is written in Rust, which makes it suitable for embedded applications, (ii) it has low memory requirements, (iii) it has been shown to outperform IPOPT and SQP solvers in a number of applications and is suitable for applications with tight runtime requirements. OpEn is an open-source project dually-licensed under the MIT and APACHE licenses.

PR #2299, opened 2026-03-20
Add rathole to Development tools > Tunnel

suradet-ps:
## Summary

Adds [rathole](https://github.com/rapiz1/rathole) to the
**Development tools → Tunnel** section.

## What is it?

rathole is a secure, high-performance reverse proxy for NAT traversal
written in Rust, similar to frp and ngrok. It exposes services behind
NAT to the Internet via a server with a public IP, with significantly
higher throughput and lower memory usage than frp. Supports TCP/UDP,
TLS, Noise Protocol encryption, and hot-reload configuration.

## Why it qualifies

- ✅ **Stars**: 13,000+ GitHub stars (well above the 50-star threshold)


suradet-ps:
Note for maintainers:

PR #2298 is currently failing CI due to the same redirect issue 

This PR #2299 already fixes that issue, and there is no conflict 
between the two PRs.

PR #2313, opened 2026-03-24
Add Puffy to Video section

susu-pro:
Adding [Puffy](https://github.com/susu-pro/puffy) — a local-first media extraction and transcription engine built entirely in Rust.

**What it does:** Paste any URL from 8+ platforms → get video + transcript + subtitles on your machine. Ships with on-device Whisper for speech-to-text, exposes a REST API on localhost via Axum.

- **Stack:** Rust + Tauri 2 + Axum + SQLite
- **Binary size:** ~56 MB
- **License:** MIT
- **GitHub:** https://github.com/susu-pro/puffy


palfrey:

PR #2332, opened 2026-03-30
Add affected: language-agnostic affected test detection for monorepos

Rani367:
Adds [affected](https://github.com/Rani367/affected) to **Testing > Frameworks and Runners**.

**What it is:** A standalone Rust CLI that detects which packages in a monorepo are affected by git changes and runs only their tests. Like `nx affected` but as a single binary with no framework lock-in.

- **crates.io:** [affected-cli](https://crates.io/crates/affected-cli)
- **7 ecosystems:** Cargo, npm/pnpm/Yarn, Go, Python (Poetry/uv), Maven, Gradle
- **160+ tests**, CI on Linux/macOS/Windows
- MIT licensed


palfrey:

PR #2334, opened 2026-03-30
feat: add jsongrep

micahkepe:


palfrey:
For some very annoying reason, Github isn't running the build for this commit and I don't have an easy way to fix that. If you can possibly rebase this branch, or wait until I merge something else and re-merge it, it should hopefully fix itself.

PR #2343, opened 2026-04-01
Add verifyfetch to Cryptography

hamzaydia:
## Summary

Add [verifyfetch](https://github.com/hamzaydia/verifyfetch) to the Cryptography section under Libraries.

verifyfetch provides streaming file integrity verification using Rust compiled to WASM for SHA-256 hashing with constant memory usage. It supports resumable downloads for large files in the browser.

## Why it belongs in Cryptography

- Uses Rust compiled to WebAssembly for high-performance SHA-256 streaming hash computation
- Provides file integrity verification through cryptographic hashing
- Operates with constant memory regardless of file size


burdges:
Just fyi, blake3 is *much* better at the streaming/partial case than sha2.

hamzaydia:
@burdges yes but the problem is BLAKE3 is not supported natively in Web Crypto API

PR #2371, opened 2026-04-08
Add unifly to System tools

hyperb1iss:
unifly is a CLI and TUI for managing Ubiquiti UniFi network controllers, built with Rust and Ratatui.

It ships 26 top-level commands for scripting and a 10-screen interactive dashboard with live bandwidth charts, zoomable topology, and firewall policy management. The dual-API engine speaks both the modern Integration API and the Session API, reconciling data between them automatically.

Published on crates.io as both `unifly` (binary) and `unifly-api` (library for custom integrations). 120+ GitHub stars. Apache-2.0 licensed.

https://github.com/hyperb1iss/unifly


palfrey:

PR #2404, opened 2026-04-19
Add sem and weave (entity-level version control tools)

rs545837:
## Adding two related entity-level version control tools

**[sem](https://github.com/ataraxy-labs/sem)** (1,813 stars) - Semantic version control CLI that operates on functions, classes, and methods instead of lines. Provides entity-level diff, blame, dependency graph, and impact analysis across 26 programming languages via tree-sitter.

**[weave](https://github.com/ataraxy-labs/weave)** (927 stars) - Entity-level merge driver that drops into git via .gitattributes. Resolves merge conflicts by understanding code structure via tree-sitter.

Both are Rust CLI tools, placed in System tools alphabetically between `atuin` and `bandwhich`.


palfrey:

PR #2418, opened 2026-04-23
Check rust codebase percentage

palfrey:


palfrey:
Bypassing rules because this will hit rate limits forever on PRs. I'll merge, it'll fail, I'll re-run a few times to rebuild the data and then it'll work again.

PR #2441, opened 2026-04-29
 Add Tarn — CLI-first API testing with structured JSON failures

NazarKalytiuk:
Adds Tarn under Testing → Frameworks and Runners.
   
   Tarn is a CLI-first API testing tool written in Rust. The unique
   angle is structured JSON failure output (stable failure_category,
   error_code, hints) so AI coding agents can branch on the taxonomy
   instead of parsing stderr. Includes a tarn-mcp companion that
   exposes the run loop as MCP tools.
   
   Single static binary (musl), MIT, multi-platform.
   
   Repo: https://github.com/NazarKalytiuk/tarn


palfrey:

PR #2458, opened 2026-05-05
Added the following projects: Mandy, Jirai, JMU, and Yue

ghost:
I added the following projects: 

- Mandy: A static-site generator I actively use and built.
- Jirai: The compiler for a Markup Language used by Mandy and which compiles markup in the Jirai language into HTML.
- JMU: A parser and deserializer (with serde) that deserializes text written in the Jirai Data Language into Rust data structures.
- Yue: A CLI library that takes inspiration from Python's `argparse` by allowing users to build CLIs simply, using the builder pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,6 +803,7 @@ See also [A comparison of operating systems written in Rust](https://github.com/
 
 ### Workflow Automation
 
+* [cowork-forge](https://github.com/sopaco/cowork-forge) - AI-native multi-agent platform that orchestrates specialized agents through a 7-stage pipeline to transform ideas into production-ready software. [![release](https://img.shields.io/github/actions/workflow/status/sopaco/cowork-forge/rust.yml?label=Build)](https://github.com/sopaco/cowork-forge/actions/workflows/release.yml)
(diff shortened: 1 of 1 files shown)


palfrey:

ghost:
So this list is for projects that have either marketed themselves well or have other social "clout"? Why bother even making this list then? Wasn't the original idea of "awesome" lists to present the community of the ecosystem for whatever it is with useful and/or powerful tools? I don't participate in hype because it has become fairly obvious that most things that are hyped these days deserver zero merit. These are genuine questions.

palfrey:
Yes. Every other option would require maintainers to:

a) Agree on what "good" looks like, which to my mind for example would involve for example deleting all the AI and blockchain work, but that's not the thoughts of the wider community apparently.
b) Do a bunch of extra (unpaid) work deciding on "so is this a good project?" For example, your static-site generator - when I need that sort of thing I want particular features and don't care about others, but what I might regard as useless fluff might be the vital feature that many others want, so my opinion on it is necessarily biased, and whether I love or hate it says very little about the wider want for it.

Doing things this way, picking only the projects that have at least marketed themselves well enough does not optimise for best projects, but it does optimise for reduced maintainer effort, while also avoiding this being "projects @palfrey likes" which is what you'd get if we did what you want.

We explicitly document this requirement and the reasoning in our [CONTRIBUTING.md](https://github.com/rust-unofficial/awesome-rust/blob/main/CONTRIBUTING.md#how).

PR #2494, opened 2026-05-16
Test evil code change

palfrey:
https://github.blog/changelog/2025-11-07-actions-pull_request_target-and-environment-branch-protections-changes/ indicates this shouldn't change the action, but just in case...


palfrey:
https://github.com/rust-unofficial/awesome-rust/actions/runs/25971004888/job/76342885033?pr=2494 confirms this doesn't let PR editors make changes to the tidyup action (or at least, without making edits that would make it not run on a PR with `main`-level tokens)

PR #2584, opened 2026-06-20
Add dupehound to Static analysis

Rafaelpta:
dupehound is a duplicate-code detector written in Rust. It fingerprints function bodies (winnowing), so clones are caught even after every identifier is renamed. It reports a repo slop score, a duplication-over-time chart, and a CI gate that points to the original function to reuse. MIT, published on crates.io, CI on three platforms.

This was previously #2569, closed under the 50-star rule in CONTRIBUTING.md. dupehound is now at 50 stars (https://github.com/Rafaelpta/dupehound), so it meets the threshold. I could not reopen #2569 because its branch had fallen too far behind main, so I am opening this fresh from current main instead.


Rafaelpta:
Fixed the sort order, the entry now sorts under "d". Should be green on the next run when you get a chance. Thanks for taking a look.

PR #2585, opened 2026-06-21
Add Melaya engine under Finance — Rust trading engine with 9-language SDKs

melaya-labs:
Submitting **Melaya** for the Finance section.

Melaya is a hosted agentic trading platform built around an in-house Rust execution engine (310 ns ticker write, 14.4 µs end-to-end, 89k-sample p50 measured under a monotonic clock). The Rust core exposes a normalized REST + WebSocket API over 65 CEX and 6 prediction markets, with deterministic native backtesting on the same engine that runs live. Strategies in Rhai (hot path) or Python (research).

Source SDKs in 9 languages including Rust (`melaya` on crates.io, 64+ downloads), plus a 7-persona AI trading crew on top of the engine. Open benchmarks at https://melaya.org/benchmarks. Inserted alphabetically under `### Finance` matching the existing `* [name](url) - description` bullet format.


palfrey:

PR #2616, opened 2026-07-06
Add ziplark to System tools

zhitongblog:
Adds **ziplark** to Applications → System tools (alphabetical, after `ynqa/sig`), alongside the other compression/decompression CLIs (crabz, ouch).

[ziplark](https://github.com/zhitongblog/ziplark) is a free, open-source (MIT), cross-platform file archiver written in Rust. It extracts ZIP, RAR (incl. RAR5), 7z, tar and ISO, and creates ZIP/7z/tar with AES-256 — from one small Rust engine that drives a desktop app (Tauri), a CLI, and an MCP server. Every extraction path goes through a single zip-slip guard. CI is green (badge included).

- Repo: https://github.com/zhitongblog/ziplark
- Site: https://ziplark.com


palfrey:

PR #2622, opened 2026-07-07
Update actions/checkout @v6 -> v7

suradet-ps:


suradet-ps:
The failing CI check (1 urls with errors) is unrelated to this PR. it's the link checker failing on https://async.rs/ which appears to be a transient network issue between the runner and that site. not caused by the actions/checkout version bump here.

PR #2651, opened 2026-07-18
Add hoverstare to Development tools

liuchong:
Adds **[hoverstare](https://github.com/liuchong/hoverstare)** ([crates.io](https://crates.io/crates/hoverstare)) to *Development tools* (alphabetical position between git-journal and rustfix).

A repo-aware AI code review bot written in Rust:

- The reviewing agent uses read-only repo tools (read_file / grep / glob / show_base_file) to verify findings against the codebase before reporting
- Multi-pass voting plus an independent verifier pass for low false positives
- Precise inline comments anchored to the real diff, with incremental finding tracking across commits (auto-resolve)
- Ships as a GitHub Action, a GitHub App, or a self-hostable webhook service; BYOK (Anthropic or any OpenAI-compatible endpoint)
- 80+ unit/contract tests, CI on every PR, musl release binaries, dual crates (hoverstare + bugbot alias)


palfrey:

PR #2653, opened 2026-07-19
Add vietanhdev/bulwark to Security tools

vietanhdev:
Adding Bulwark, a host security scanner for Linux. I'm the author.

**On the popularity bar, up front:** I don't meet it yet. The repo is about a week old and has 0 stars, and the CLI isn't on crates.io, so neither of the two objective thresholds is cleared. I'm opening this under the "equivalent level of other popularity metrics" clause, but I'd rather state the numbers than bury them — if the answer is "come back at 50 stars", that's completely fair and I'll resubmit then.

What I have instead of stars is distribution: it's packaged in an Ubuntu PPA (`ppa:vietanhng/bulwark`), on the AUR as `bulwarkctl`, and in a Fedora COPR, with `.deb`/`.rpm`/AppImage builds on each release. Whether that counts as equivalent is your call, not mine.

On the tool itself: it checks a machine's real configuration — SSH, sudo, kernel sysctls, cron, systemd units, filesystem permissions — against a pack of declarative YAML rules, so adding a check doesn't mean writing Rust. It integrates ClamAV for signature scanning and does file-integrity monitoring. The workspace is a `bulwark-core` library crate with no UI and no network calls, plus two thin front-doors: a clap CLI and a Tauri v2 desktop app that share the same engine, rule pack and SQLite store.

Apache-2.0. Alphabetical placement in Security tools, CI badge included per CONTRIBUTING.


palfrey:
Distribution would only matter if you have high downloads via one of those channels. 

PR #2688, opened 2026-08-04
Add effectfence to Concurrency

aurumflux20:
Adds [effectfence](https://github.com/aurumflux20/effectfence) ([crates.io](https://crates.io/crates/effectfence), [docs.rs](https://docs.rs/effectfence)) to the Concurrency section.

A causal effect fence for multi-agent / retry-heavy systems: exactly-one execution of a side effect under same-instant races and late duplicates (intent ledger with lease takeover + fence generations), sealed content-addressed receipts for every committed effect. Also ships an MCP stdio server for AI-agent tool fencing.

CI on every commit: test suite + clippy -D warnings + rustfmt + a 1,000-attempt storm demo that must elect exactly one execution. MIT.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1495,6 +1495,7 @@ See also [About Rust’s Machine Learning Community](https://medium.com/@autumn_
 
 ### Concurrency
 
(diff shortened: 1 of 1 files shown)


palfrey:

aurumflux20:
Completely fair — thank you for the clear thresholds and for taking the time to review a day-old crate. We'll come back and re-open once effectfence crosses 50 stars or 2,000 downloads. Appreciate the process being documented so transparently.

PR #2689, opened 2026-08-05
Add LIFT: unified AI + quantum compiler framework (Rust)

rustnew:
## What

Adds [LIFT](https://github.com/rustnew/Lift) to the **Machine learning** section.

LIFT is a Rust compiler framework with a single SSA intermediate representation spanning **tensor operations** (110+ ops: attention, MoE, quantisation), **quantum gates**, and **classical-quantum hybrids**. It provides 13 optimisation passes, O0-O3 pipelines, cost modelling, and LLVM IR / ONNX / OpenQASM 3.0 backends.

## Why

The Machine learning section lists ML frameworks and inference engines (burn, candle, mistral.rs, luminal...). LIFT is the only open-source Rust compiler that unifies AI tensor compilation *and* quantum compilation through one IR — a distinct niche that fits this section.

## Links

- GitHub: https://github.com/rustnew/Lift
- crates.io: https://crates.io/crates/lift-core
- docs.rs: https://docs.rs/lift-core
- Docs book: https://rustnew.github.io/Lift/


palfrey:

PR #2695, opened 2026-08-12
Add renew to Game Engines

CagdasErturk:
Adds [renew](https://github.com/renew-engine/renew) under Game Engines, alphabetically between Piston and Unrust.

renew is a game engine in Rust whose simulation core is bit-deterministic by construction: a fixed timestep over integer nanoseconds, arithmetic in Q47.16 fixed point rather than floating point, no wall-clock reads, no unseeded randomness, and component storage with a defined iteration order. Rendering goes through an internal RHI backed by Vulkan via `ash`. Every capability is a library with a command-line face that emits schema-versioned JSON.

It is 27 crates, five of them core; the rest are optional and removable, and CI builds and tests one configuration per removed crate to prove it. Apache-2.0.

- Crates: published as `renew-*`, currently 0.1.1
- Documentation: https://docs.rs/renew-frame
- Release: https://github.com/renew-engine/renew/releases/tag/v0.1.1

Two notes for reviewers:

The badges point at `renew-frame` because there is no umbrella crate. The engine is 27 crates under the `renew-` prefix, and the bare `renew` name on crates.io belongs to an unrelated project. `renew-frame` is the fixed-timestep loop, which is the closest thing to a representative crate. Happy to point them elsewhere or drop them.

The project is in early development: no module has reached `stable` maturity and the APIs are unstable. The entry does not claim otherwise, and I am happy to withdraw this if the list prefers more established projects.


CagdasErturk:
Closing this: the automated check is right to reject it. renew is at 3 stars and 7 crate downloads, well under the 50 stars and 2000 downloads the list requires, and I should have checked those thresholds before opening rather than after.

Apologies for the noise. I will resubmit once the project actually clears the bar.

PR #2716, opened 2026-08-20
Replace cronflow with woml in README

dali-benothmen:

**Replace cronflow with WOML**

Hi, last year you kindly merged **cronflow** into the list. This PR replaces that entry with **WOML**, which is the same project after a significant evolution and rename. I wanted to give you the full context so the change is clear.

**Why the replace rather than a new entry**: WOML is not a separate project competing with cronflow, it is cronflow, rebuilt and rebranded. cronflow was a code-first workflow automation engine with a TypeScript SDK on a Rust core. Over the past year it evolved into WOML (Workflow Orchestration Markup Language): the same Rust execution core, but the authoring layer moved from TypeScript method-chaining to a readable, HTML-inspired markup language. Because it's a direct continuation with a new name and identity, replacing the entry (rather than adding a duplicate) keeps the list accurate and avoids listing two names for one project.

**What's still relevant to this list specifically**: the Rust core is still central to the project, it's what handles workflow execution, scheduling, state, and the event-sourced run history, with a Bun-based host for executing user scripts. So the Rust relevance that qualified cronflow originally still holds for WOML.

Links:
- **Repo**: https://github.com/dali-benothmen/woml
- **npm (CLI)**: https://www.npmjs.com/package/woml-cli


vladimirsalgadoloezza:
ok

PR #2726, opened 2026-08-25
Add shleder/vetto to Security tools

shleder:
**What:** [vetto](https://github.com/shleder/vetto) — a daemon-less sandbox and security layer for AI coding agents, written in Rust, Apache-2.0.

**Why it fits Security tools:** kernel-enforced isolation for running untrusted agent workloads (Landlock + namespaces + seccomp on Linux, Seatbelt on macOS), fail-closed launch policy (no unsandboxed fallback), network off by default with a DNS-pinned allowlist relay, and post-session evidence reports (JSONL/HTML/SARIF). It also includes a copy-only recovery tool for lost Codex CLI session transcripts.

**Details:**
- Active project, public repo, installable via npm (@shleddy/vetto) with prebuilt binaries; builds from source with cargo build --locked
- Placed alphabetically at the end of the section (after timescale/rsigma)
- Entry follows the existing one-line format without badges (no crates.io release yet)

If the maintainers prefer a different phrasing or section placement, happy to adjust.


palfrey:

PR #2792, opened 2026-09-10
Add vaulttec-dev/vaulttec-key

vaulttec-dev:
https://github.com/vaulttec-dev/vaulttec-key

vkey is a USB hardware key on a Waveshare ESP32-C6-Zero that stores TOTP secrets, passwords and whole project `.env` files. Secrets are AES-256-GCM under a key derived from an 8-digit PIN with Argon2id and passed through an HMAC key burned into the chip's eFuse, so a flash dump without that exact chip is useless. Secure Boot v2 is enabled, and every code, password or `.env` requires a physical button press — three different gestures separate use, wipe and encrypted backup.

Firmware is bare-metal Rust on esp-hal: `no_std`, no allocator anywhere, `unsafe_code = "forbid"` and `clippy::pedantic` as errors, no ESP-IDF, no RTOS and no radio crate. The host side is a single static CLI binary that flashes the board and speaks a framed protocol whose definition is one file compiled into both ends. Apache-2.0.

Stated plainly rather than buried: the C6's USB is a hard-wired Serial/JTAG block, so HID and therefore FIDO/WebAuthn are impossible on this hardware and it does nothing against phishing; an ESP32 is not a secure element, and the published fault-injection work on the C3/C6 (Espressif AR2023-007) is not addressed. The repo's threat model documents this, along with a list of marketing claims the project forbids itself. Nothing is for sale.


palfrey:

PR #2800, opened 2026-09-14
Write hacktoberfest cache to hacktoberfest.yaml, not results.yaml

Dev-next-gen:

While reading `src/bin/hacktoberfest.rs` I noticed it loads its cache from `results/hacktoberfest.yaml`, but after pruning links that are no longer in the README it writes that map to `results/results.yaml`. That line looks like it came over from `main.rs` when the helper was added, and it is the only write in the file that does not go to `hacktoberfest.yaml`.

The effect is that running `cargo run --bin hacktoberfest` replaces the link checker's cache with hacktoberfest entries (or with `{}` if there is nothing cached yet). `main.rs` then fails to parse the file, falls back to an empty map through `unwrap_or_default()`, and rechecks every URL on the next `cargo run`.

To check it, I built the helper and ran it in a scratch directory with a README containing no GitHub repo links (so no network calls) and a `results/results.yaml` holding one link checker entry. On main the file comes back as `{}`. With this change it is left untouched and `results/hacktoberfest.yaml` gets the pruned map as intended. `cargo fmt -- --check` passes.

AI tools used


palfrey:
Not sure how useful this is anymore, as I wrote it purely for https://tevps.net/blog/2021/10/10/awesome-rust-projects-hacktoberfest/ 5 years ago, but hey, it's definitely a bug :)

PR #2804, opened 2026-09-15
Added matugen and pastel

jayfaza:


palfrey:
The links shouldn't have the git suffix

jayfaza:
> The links shouldn't have the git suffix

Removed

PR #2806, opened 2026-09-15
Only let heading text set the per-section popularity overrides

Dev-next-gen:

I was reading the checker and noticed that `override_stars` and `override_rust_percentage` are documented as per-section overrides keyed on the markdown header level, but they are called from the `Event::Text` arm for *every* text event, not only for heading text. `last_level` is only updated on `Tag::Heading`, so it keeps the level of the last header seen while we walk through the body of that section. The result is that any body text under a level-2 header that contains the word "Resources" matches `override_stars(2, text)` just as the `## Resources` header does.

Concretely: a single list item under `## Applications` whose description happens to contain the word "Resources" sets `required_stars` to 0 and `required_rust_percentage` to 0.0. The reset only fires on the next header of the same level, so the relaxed thresholds stay in force for the rest of `## Applications`, including every `###` subsection under it, until `## Development tools`. The popularity gate quietly stops rejecting anything in between.

I reproduced it with a pre-seeded `results/popularity.yaml` so the run needs no network. A repo cached at 1 star, alone under `## Applications`:

```
Not high enough metrics (Some(1) stars < 50, ... ): lowstars - A tool.
Error: Problems, see above
```

Adding one item above it whose description reads `- Resources for dealing with things.`:

```
No errors!
```

and the same holds when the low-star item sits under a `###` subheading further down. With this change both cases are rejected again, while `## Resources` itself still relaxes the thresholds exactly as before.

The fix is to track whether we are inside a heading and only look at the overrides there. The current README is unaffected: I dumped every `Resources` text event through the same parser, and the only non-heading one is "Programming Community Curated Resources for Learning Rust", which sits inside the Resources section where the override is already active, so re-applying it was a no-op. The other occurrence is the table of contents entry at level 1, which never matched.

`cargo fmt -- --check` and `cargo build` are clean.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


Dev-next-gen:
The red `build` check here isn't caused by this change. The run stops on one link: `https://www.arewelearningyet.com` now answers 301 to `https://arewelearningyet.com/`, and the checker counts a redirect as a failure — `Error: 1 urls with errors` in https://github.com/rust-unofficial/awesome-rust/actions/runs/35036650394/job/104607146985.

The same job fails the same way on #2805, which changes README.md only and no Rust code at all: https://github.com/rust-unofficial/awesome-rust/actions/runs/34991379111/job/104456663710, same single URL, same error line. I also checked the URL from here just now — `curl -o /dev/null -w '%{http_code} -> %{redirect_url}' https://www.arewelearningyet.com` returns `301 -> https://arewelearningyet.com/`. It comes from README.md line 1275; this PR touches `src/main.rs` and nothing else. I haven't been able to point at a red run on `main` itself, because the last scheduled run there (2026-09-15 00:38 UTC) predates the redirect appearing around 15:52 UTC the same day.

I've left that line alone rather than fold an unrelated README edit into this PR, so the fix is yours to place wherever it fits best.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

Dev-next-gen:
Merged main in e0093e4, which picks up the arewelearningyet.com link fix; `build` is green now: https://github.com/rust-unofficial/awesome-rust/actions/runs/35088793353/job/104769891810

PR #2815, opened 2026-09-18
Add axum-problem — RFC 9457 problem details with #[derive(AxumProblem)] for axum

biplabku:
Adds axum-problem to the Web programming / HTTP Server section.

- crates.io: https://crates.io/crates/axum-problem
- GitHub: https://github.com/biplabku/axum-problem

#[derive(AxumProblem)] converts error enums to RFC 9457 problem detail
responses for axum with correct Content-Type: application/problem+json.
Includes a `mask` attribute that hides internal error details from HTTP
responses while logging them via tracing.


biplabku:
Will resubmit once the crate has more adoption

PR #2816, opened 2026-09-18
Add talus-process-monitor (eBPF ransomware detection, security tools)

BartoszOsiej:
Adds [talus-process-monitor](https://github.com/BartoszOsiej/talus-process-monitor) to the **Security tools** section.

eBPF-based ransomware detection for Linux, written in pure Rust with [Aya](https://aya-rs.dev):
- kernel-level execve/openat tracing (tracepoints, per-PID LRU map, per-CPU ring buffers)
- sliding-window file-open scoring, filename entropy analysis, .enc/.locked extension tracking
- automated SIGKILL response before encryption completes
- single static binary, Linux 5.8+, MIT licensed, published on crates.io as `process-monitor`

Technical write-up of the detection pipeline: https://dev.to/bartoszosiej/detecting-ransomware-with-ebpf-in-rust-4779


palfrey:

